PR #10, opened 2017-06-03
Refactor classes

xadahiya:
### Changes 
- Refactored classes.py to generate classes directly from the subsystems and spacecraft json-ld

- Add subsystem Vocab from http://ontology.projectchronos.eu/subsystems/?format=jsonld

- Add spacecraft Vocab from http://ontology.projectchronos.eu/spacecraft/?format=jsonld


chrizandr:
:+1

PR #14, opened 2017-06-06
Views

xadahiya:
- Add Initial Views template and IRI structure
- Auto Data generation and insertion in PSQL (for COTS)
- Directory Restructure
- Minor bug fixes 
- @chrizandr PR #13 Merge 
- CRUD operations using sqlalchemy


requirements.txt
```diff
@@ -1,14 +1,21 @@
...
 persisting-theory==0.2.1
+psycopg2==2.7.1
 pyparsing==2.2.0
+python-dateutil==2.6.0
+pytz==2017.2
 six==1.10.0
-SQLAlchemy==1.1.10
 thespian==3.5.2
-Werkzeug==0.12.2
+wsgiref==0.1.2
```
Mec-iS: BUG: there is not Python 3 version for `wsgiref` because it has been merged in the STDLIB as `http.server`, use `http.server` instead.

PR #22, opened 2017-06-11
Develop

xadahiya:
-Item Collections done
- Automation using docker to setup demo server
- Crud fix
- Tests for crud.py
- Generator.py refactor
- Insert_data refactor
- Test directory refactor
- Minor import fixes
- Add Server test template
- Api built with dynamic context generation


chrizandr:
You can't insert any other type of subsystem. That is because the other classes are not yet in the Database. They still need to be added.
And we still need to decide how we want to automatically add them
I added thermal manually. I forgot to dereference it. But for now you will have to add all classes and properties to RDFClass and BaseProperty manually. I have inserted the properties used by the random generated data. But there are others that need to be added
There is no issue with the insert method, only that the relevant classes are not present in the DB

Also, about the type_ attribute, I was wondering if it is needed at all since we can add a property called `type` and store that in graph itself. For now I will add the type and the classes. I will dereference as well.

hydrus/data/crud.py
```diff
@@ -56,6 +56,10 @@ def insert(object_, id_=None):
     """Insert an object to database [POST] and returns the inserted object."""
     # NOTE: We are inserting the object, no need to check if similar one already exists.
     #       Data can be redundant/identical, they must have different "@id"
+
```
chrizandr: You can't insert any other type of subsystem. That is because the other classes are not yet in the Database. They still need to be added.
And we still need to decide how we want to automatically add them
I added thermal manually. I forgot to dereference it. But for now you will have to add all classes and properties to RDFClass and BaseProperty manually. I have inserted the properties used by the random generated data. But there are others that need to be added
There is no issue with the insert method, only that the relevant classes are not present in the DB

Also, about the type_ attribute, I was wondering if it is needed at all since we can add a property called type and store that in graph itself. For now I will add the type and the classes. I will dereference as well.

hydrus/data/crud.py
```diff
@@ -56,6 +56,10 @@ def insert(object_, id_=None):
     """Insert an object to database [POST] and returns the inserted object."""
     # NOTE: We are inserting the object, no need to check if similar one already exists.
     #       Data can be redundant/identical, they must have different "@id"
+
+    # TODO: @chrizandr You're not setting instance.type_ and you're not using keymap
+    # to dereference class types. Try any other category other then thermal for insertion
+    # Also, make sure crud.py passes all the tests in test_db.py (Currently all failing)
```
chrizandr: You can't insert any other type of subsystem. That is because the other classes are not yet in the Database. They still need to be added.
And we still need to decide how we want to automatically add them
I added thermal manually. I forgot to dereference it. But for now you will have to add all classes and properties to RDFClass and BaseProperty manually. I have inserted the properties used by the random generated data. But there are others that need to be added
There is no issue with the insert method, only that the relevant classes are not present in the DB

Also, about the type_ attribute, I was wondering if it is needed at all since we can add a property called type and store that in graph itself. For now I will add the type and the classes. I will dereference as well.

xadahiya: @chrizandr aren't we using insert_classes.py to automatically insert classes to the Database. I think the work flow is we run `db_models.py` to setup database then we add classes using `insert_classes.py` and then we can insert data using `insert_data.py` which uses gemerator.py to automatically generated and insert data. Only the properties are automatically inserted during runtime. What say?

xadahiya: @chrizandr I know we have `type` attribute in our Graph but that is for what type of graph it is. For example `III` or IAI` etc. But the type attribute in `instance` is for which `class` the object belongs to. For example `Spacecraft_power` or `Spacecraft_CDC` etc. The both are entirely different cases and are needed to be defined separately.

chrizandr: You asked for `instance.type_`, the `Graph.type_` attribute is not for the anyone to modify. The ORM handles it in the back. What is your doubt?

xadahiya: You're partially right. We need to tell which class an instance belongs to but I don't think there's a need to use a `GraphIAC` triple to store that. Also we have `GraphIII` and `GraphIIT` where where we'll have to define which class every instance belongs to. We already have a system defined for this in the Instance table.
```python
class Instance(Base):
    """Model for Object/Resource.
    Instances are instances of some kind/classes that are served through the API.
    """

    __tablename__ = "instances"

    id = Column(Integer, primary_key=True)
    name = Column(String)
    type_ = Column(Integer, ForeignKey("classes.id"), nullable=True)
```
The `type_` attribute can be used to define which class any instance belongs to. What say?

Issue #17, opened 2017-06-11
Start writing a "mini-wiki" in the README


Mec-iS:
Write in the README how to run the example to dump the data and then run the server. Explaining which logic is followed to define the API from the RDF vocabulary to the endpoints.

PR #23, opened 2017-06-14
Wiki

chrizandr:
Wiki as required in #17


Mec-iS:
there are still some merging headers in the code.

chrizandr:
Updates added directly to the wiki section

PR #24, opened 2017-06-19
Develop

chrizandr:
Keeping this here so both can merge their commits easily.


xadahiya:
Same as #22

chrizandr:
It's kept here so that's it's easier for you to know when there are updates on my fork and vice versa

PR #36, opened 2017-07-09
Feature/add simulation classes

xadahiya:
- Status Codes fix (modified all status codes to match [W3C standards](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html), hydrus should work with all generic clients now).
- Drone, Server parsed classes add
- Bug pointed by Markus Fixed
- Minor bug fixes


xadahiya:
used in https://github.com/HTTP-APIs/hydrus/pull/37

PR #37, opened 2017-07-15
Fix/crud

xadahiya:
- Crud.py fix, now supports every type of objects
- Tests updated
- Status Codes fix (modified all status codes to match [W3C standards](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html), hydrus should work with all generic clients now).
- Drone, Server parsed classes add
- Bug pointed by Markus Fixed
- Minor bug fixes


chrizandr:
Merged into #38

PR #38, opened 2017-07-15
Added the doc writer to Hydrus

chrizandr:
Everything seems to be working for now. 

For @xadahiya 
- Made a few changes to crud also
- Removed the scripts for those generators, everything is inside the doc_writer now
- Also tried to make endpoints dynamic, instead of being /api/... it is /<API_NAME>/..., same thing is passed to generate the doc.
- See the different classes in app, you'll see how API_DOC works.
- Check `hydrus.metadata.server_doc_gen` to see how to change the doc.
- Didn't try it with HydraConsole, leave that to you
- Add classes using `hydrus.data.doc_parse`. 
- Make changes to the `db_models` and remove `Instance.name`

All the best :smile:


chrizandr:
Switched to develop #39

PR #40, opened 2017-07-16
Server bug fix

xadahiya:
- Server bug fixed, now working with hydra console
- Tests updated
#will finish drone once we finalize the design.

### DIFF
--- a/hydrus/hydraspec/doc_writer.py
+++ b/hydrus/hydraspec/doc_writer.py
@@ -302,7 +302,6 @@ class Context():
                 "supportedClass": "hydra:supportedClass",
                 "supportedProperty": "hydra:supportedProperty",
                 "supportedOperation": "hydra:supportedOperation",
-                "code": "hydra:statusCode",
                 "statusCodes": "hydra:statusCodes",
                 "label": "rdfs:label",
                 "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
--- a/hydrus/hydraspec/doc_writer.py
+++ b/hydrus/hydraspec/doc_writer.py
@@ -68,9 +68,7 @@ class HydraDoc():
             "@context": self.context.generate(),
             "@id": self.base_url + self.API + "/vocab",
             "@type": "ApiDocumentation",
-            "title": self.title,
             "description": self.desc,
-            "entrypoint": '/'+self.entrypoint.api,
             "supportedClass": [x.generate() for x in parsed_classes + self.other_classes + collections + [self.entrypoint]],
             "possibleStatus": [x.generate() for x in self.status]
         }
@@ -133,8 +131,8 @@ class HydraClassProp():
           "title": self.title,
           "property": self.prop,
           "required": self.required,
-          "readable": self.read,
-          "writeable": self.write
+          "readonly": self.read,
+          "writeonly": self.write
         }
         return prop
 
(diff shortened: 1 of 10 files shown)


hydrus/metadata/drone_doc_gen.py
```diff
@@ -14,15 +14,15 @@ def drone_doc(API, BASE_URL):
                        BASE_URL)
 
     # Status Class
-    status = HydraClass("Status", "Status", "Class for drone status objects")
+    status = HydraClass("http://hydrus.com/Status", "Status", "Class for drone status objects")
```
chrizandr: Why do we need the http://hydrus.com/ prefix? Status is defined in the API doc. Why have it dereference to a non existent URI? 
vocab:Status should be fine I think.

hydrus/hydraspec/doc_writer.py
```diff
@@ -377,20 +377,7 @@ def generate(self):
...
-                        "@id": "_:%s_collection_retrieve" % (self.name.lower(),),
-                        "@type": "hydra:Operation",
-                        "method": "GET",
-                        "label": "Retrieves all %s entities" % (self.name,),
-                        "description": None,
-                        "expects": None,
-                        "returns": "vocab:%sCollection" % (self.name,),
-                        "statusCodes": [
-                        ]
-                    }
```
chrizandr: Collections should support GET by default. Why is it removed?

hydrus/hydraspec/doc_writer.py
```diff
@@ -304,6 +303,7 @@ def __init__(self, address, adders={}, class_=None, collection=None, entrypoint=
                 "supportedProperty": "hydra:supportedProperty",
                 "supportedOperation": "hydra:supportedOperation",
```
chrizandr: Remove "code" it is redundant

hydrus/metadata/drone_doc_gen.py
xadahiya: We need a placeholder url to be able to use hydra-py. It uses a schema namespace to find operations we can't do that with vocab: status. Same goes for all.

hydrus/hydraspec/doc_writer.py
xadahiya: Okay

hydrus/hydraspec/doc_writer.py
xadahiya: SupportedOperations are already defined at vocab:<collection_name>, it'll just retrieve SupportedOperations from there. Also collections support both get and post. We should be able to add new items to the collection ( See event API example)

chrizandr: As discussed in the mailing list, the use of Collection or Class for creating new object is subjective. But the GET operation is inherently supported. APIs may use api/class/ to create objects instead of api/classCollection.
But it's fine, we'll have to modify the operations in the Collection then

hydrus/metadata/drone_doc_gen.py
chrizandr: The vocab has an `@id: <server_url>/api/vocab` we can use that to replace `vocab:`?

xadahiya: The way hydra-py finds resources is say we have a SCHEMA `http://schema.org` then `SCHEMA. addAction` will look for schema.org/addAction . I don't see any problem using a placeholder

hydrus/hydraspec/doc_writer.py
xadahiya: It think collections should be used to add new resources and `/id` can be used to modify or delete resources. What say?

hydrus/hydraspec/doc_writer.py
xadahiya: done

hydrus/hydraspec/doc_writer.py
chrizandr: That has to be specified in the APIDoc, we cannot decide that. I'll add some things to the writer for checking that.

hydrus/metadata/drone_doc_gen.py
chrizandr: There is no problem if the links are not dereferenced. Wanted to know why we couldn't use the actual URI, since it can be dereferenced as well. It's fine

PR #41, opened 2017-07-18
Containerization 

xadahiya:
- Switched to docker-compose v2
- Removed Simulation Docs ( will be added from simulation)
- Better Containerization
- Dynamic Db selection ( psql/sqlite) based on DB_URL Environment variable


chrizandr:
Added to #42

PR #51, opened 2017-07-24
New doc_maker and pluggable docs and connectors

chrizandr:
This PR contains:
- Hydrus is made to function as a package. All we need to do is plug in the database and API Doc and the server should be able to run without modifying any code.
- A script called `main.py` is written as a sample script used to plug in the necessary objects and run hydrus.
- A new `doc_maker` is added that can convert an API Documentation[written in a Python dict] to `doc_writer` objects as defined below:
doc_writer

Todo:
Add the new changes and descriptions in the wiki


xadahiya:
Looks good 👍

PR #52, opened 2017-07-25
Bugs fixes

xadahiya:
- HYDRUS_SERVER_URL context fixed
- Docs updated
- Update setup requirements


hydrus/tests/test_app.py
```diff
@@ -7,7 +7,7 @@
 import os
```
chrizandr: Great job here. I forgot to update this :)

PR #53, opened 2017-07-26
Develop

chrizandr:
This PR includes:
- Removed settings.py and db_init.py
- Added an app_factory method to create multiple APIs using the same source
- Made the server name pluggable as well
- Updated test_app and test_crud, all passing successfully.
- Made session as a neccessary argument in all crud operations
- Modified db_models to use a local database, since there is no settings.py for DB_URL

Todo:
Update the docs
Update wiki


chrizandr:
Will get it done :+1:

Mec-iS:
@chrizandr 
A quite important thing, how error messages are returned.
In general, the application should return HTTP responses only from the Flask views. So every errors that may happen like:
```
def get(id_, type_, api_name, session, recursive=False):
    """Retrieve an Instance with given ID from the database [GET]."""
    object_template = {
        "@type": "",
    }
    try:
        rdf_class = session.query(RDFClass).filter(
            RDFClass.name == type_).one()
    except NoResultFound:
        return {400: "The class %s is not a valid/defined RDFClass" % type_}
```
it should not return a dictionary BUT `raise NoResultFound('The class %s is not a valid/defined RDFClass" % type_')`, then at view level you catch the exception and return the right HTTP response (using pseudo-code):
```
try:
   get()
except NoResultFoundError as e:
   return JSONResponse({error: true, message: e}, status=400)
except SomeOtherError as e:
    return JSONResponse({error: true, message: e}, status=404)
```
The only "layer" allowed to return to the user is the view layer. Anything below it should raise, and the exception caught and returned by the view layer. Letting the logic methods to return an error message makes the stack not consistent as the only part of the stack allowed to return a message are the views.
For example, as some lines of code in the view are not allowed to call directly an SQL query (because there is the DB layer below to do that), a logic method cannot return directly an error message because it is task of the view to talk to the client. It violates the separation of concerns.

chrizandr:
I see, wil add some custom errors that can be caught and turned into proper HTTP codes. Thanks :)

xadahiya:
Merging, as @chrizandr said he'll creating a new PR for more changes.

PR #75, opened 2018-02-13
Converted directories to package

### LINKED ISSUE #74: Error while Setup the Project
The Project is using some files from inside a package directory, but that directory is not listed as a package due to which it is throwing import error while executing the `main.py` script.

OS : Debian 9
Python 2.7

Screenshot : 
image

yashLadha:
This PR is to resolve the `Import Error` issue in setting up of the project.

Closes #74


yashLadha:
Ping @chrizandr @Mec-iS

chrizandr:
It is good to have separate __init__ files. 
However, I'm doubtful if this has anything to do with import, since we always treat Hydrus as a package and import using the complete paths. There may be some other reason you are unable to import, I suggest you open an issue, specifying exactly what problem you are facing.

PR #78, opened 2018-02-13
Sync develop with master

xadahiya:


chrizandr:
Do we still need the develop branch?

xadahiya:
I think it would be better if we work on develop and then merge with master
after some significant changes. Lorenzo ?

chrizandr:
NVM, I didn't see the chat on Gitter. 😛

Issue #80, opened 2018-02-13
Server is not stoppable


nityanandagohain:
Currently we are using serve_forever() which gives following error when it get's a keyboard interrupt.

![screenshot from 2018-02-13 16-03-36](https://user-images.githubusercontent.com/26831659/36146417-93f5f1e6-10da-11e8-8081-c2de2a7828e3.png)

I want to work on this.
Some help will be very much appreciated.

xadahiya:
Fixed in #81

PR #81, opened 2018-02-13
KeyboardInterrupt Added

ayushiiit:
Added keyboard interrupt using try catch so that server will stop on the interruption by keyboard. Refer issue #80 .


Mec-iS:
Thanks for the PR.
Any change should be first PRed first in `develop`, `master` can only receive merge from `develop`.

ayushiiit:
Ok Done. Thanks !

Mec-iS:
Please confirm also that all tests pass successfully.

Mec-iS:
Reference #83

chrizandr:
Also to be noted that this is not really an issue. A keyboard interrupt is a way for us to stop the server. Although this is not the only way to do it. You could just use the built in kill command on the OS which would send a sigkill to the process. 

A server is supposed to continuously listen for requests and so it needs to be always running. If we actually have some other use of keyboard interrupts, then it would make sense to catch them and do some operations. However, since there is no explicit operation, not catching the exception is the same as catching it and doing nothing (which is what you did in your code)

I appreciate you trying to make changes to the server, nice job on that part. I will merge these changes still, but please make sure that you're not creating issues just for the sake of getting PRs accepted.

PR #91, opened 2018-02-16
Update README.md

### LINKED ISSUE #90: Instructions to use virtual environment missing in the README
The dependencies may conflict with any other modules presently installed in the user's machine. It's better to use a virtual environment to isolate development environments.

ashwani99:
Add instructions to use a virtual environment
Fixes #90

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,13 @@ cd hydrus
 git checkout -b develop origin/develop
 ```
 
+Create a virtual environment and activate it:
+```bash
+virtualenv venv --python=`which python3`
+
+source venv/bin/activate
+```
+
 Install requirements and run the `main.py` script:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,6 +47,13 @@ cd hydrus
 git checkout -b develop origin/develop
 ```
 
+Create a virtual environment and activate it:
+```bash
+virtualenv venv --python=`which python3`
```
Mec-iS: as far as I remember, this requires `virtualenv` to be installed in the system.
In Python 3 the easiest way of create a venv is `python3 -m venv env_name`

ashwani99: @Mec-iS Yes, you're right. Thanks for correcting 👍

Issue #60, opened 2017-08-04
Implement type annotations using mypy


Mec-iS:
Try to refactor the codebase by implementing type annotations and type checking before runtime using `mypy`

aswinzz:
can i work on this issue?

Mec-iS:
Sure. Accept the invitation to the Github team and [follow the guidelines](https://github.com/HTTP-APIs/hydrus/wiki/Starting-Material)

vddesai1998:
Can I help with this issue if no one is working on it?

Mec-iS:
In general,anybody can fork the repo and start working on the fork. Once you have something done we can consider opening a branch in which different people working on the same project can collaborate.

Mec-iS:
Merged #96

vddesai1998:
Hey @Mec-iS  why type annotations from app.py are removed?
Should I open a PR type annotating app.py?
I am not opening a issue because we already has this issue.

xadahiya:
@vddesai1998  See #128

PR #96, opened 2018-02-17
Implement type annotations using mypy

vddesai1998:
Closes https://github.com/HTTP-APIs/hydrus/issues/60

### Checklist
- [x] I have created/claimed the issue that this PR resolves.


hydrus/app.py
```diff
@@ -1,411 +1,434 @@
...
+                    except Exception as e:
+                        status_code, message = e.get_HTTP()
+                        return set_response_headers(jsonify(message), status_code=status_code)
+
+            return set_response_headers(jsonify({400: "Data is not valid"}), status_code=400)
+
+        abort(405)
+
+    def put(self, id_, type_):
+        """Add new object_ optional <id_> parameter using HTTP PUT."""
```
chrizandr: I think this was missed

vddesai1998:
How can I resolve this conflict?

vddesai1998:
Tests were passing locally but once I made my branch up to date with develop authorization tests are failing

examples/drones/doc_gen.py
```diff
@@ -3,8 +3,8 @@
 from hydrus.hydraspec.doc_writer import HydraDoc, HydraClass, HydraClassProp, HydraClassOp
 import json
 
-
 def doc_gen(API, BASE_URL):
+    # type: (str, str) -> HydraDoc
```
xadahiya: The type notations look fine but currently, these are nothing but comments. You need to implement static type checking and add mypy to the testing environment. 
See https://medium.com/@ageitgey/learn-how-to-use-static-type-checking-in-python-3-6-in-10-minutes-12c86d72677b and http://mypy.readthedocs.io/en/stable/getting_started.html for more info.

hydrus/data/db_models.py
```diff
@@ -3,11 +3,12 @@
 from sqlalchemy import create_engine, ForeignKey
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy import Column, Integer, String
+from typing import Any
```
xadahiya: Add typing to requirements.txt

examples/drones/doc_gen.py
xadahiya: The type notations look fine but currently, these are nothing but comments. You need to implement static type checking and add mypy to the testing environment. 
See https://medium.com/@ageitgey/learn-how-to-use-static-type-checking-in-python-3-6-in-10-minutes-12c86d72677b and http://mypy.readthedocs.io/en/stable/getting_started.html for more info.

hydrus/data/legacy/generator.py
```diff
@@ -1,6 +1,10 @@
 """Generate random objects for SubSystem classes."""
 
 import random
+from pyannotate_runtime import collect_types
```
xadahiya: Add pyannotate to requirements.txt

hydrus/tests/__init__.py
```diff
@@ -1 +0,0 @@
-
```
xadahiya: Write tests for Static Type Checking using mypy.

vddesai1998:
Thanks for reviewing. I will start implementing changes soon...

vddesai1998:
And these are not mere comments but way to annotate types in mypy which is compatible with both python 2 and python 3. And pyannotate is not needed as requirement in our codebase. I used that open source tool to generate type annotations for our codebase . From now on contributors will annotate types while implementing new code. 

I will replace  those comment based annotations with python3.5 annotations syntax and write tests.

xadahiya:
1. `And these are not mere comments but way to annotate types in mypy which is compatible with both python 2 and python 3` : As @Mec-iS  already specified, `As a generic rule, everything should work and be tested for Python 3.5.2 and above`. You don't need to make the codebase compatible with both python2 and python3.
2. `pyannotate is not needed as requirement in our codebase` : Doesn't matter if we need it in our codebase or not, if you're importing the module and it's not installed it will show an error. You should remove the import line if it's not required any more.

vddesai1998:
Yeah I missed that. I should remove that unused import. Again thanks for reviewing

Mec-iS:
As specified by @xadahiya we want the types annotations to be in the code not in comments.

vddesai1998:
@Mec-iS @xadahiya Can you review it ? I made suggested changes , removed unused imports and  used python3 annotation syntax.

examples/drones/doc_gen.py
vddesai1998: How should I add mypy to testing environment? As of now, I am running mypy in  hydrus directory with command  ```mypy  .``` to check the type errors. 
Tried mypy-pytest tool but it was too slow.
I think the best way would be to run mypy in the root directory as I am doing before pushing any changes.

Mec-iS:
Please confirm that you run the `mypy` command and what it returns

vddesai1998:
It returns errors about missing stub files. I tried to solve this but stub files for few libraries are not included in typeshed.
But when I run ```mypy --ignore-missing-imports``` on individual files it does not shows any errors.
I also ran ```mypy  .  --ignore-missing-imports``` in hydrus directory.

hydrus/app.py
```diff
@@ -1,443 +1,444 @@
...
-        if get_authentication():
-            if request.authorization is None:
-                return failed_authentication()
-            else:
-                try:
-                    auth = check_authorization(request, get_session())
-                    if auth is False:
-                        return failed_authentication()
-                except Exception as e:
-                    status_code, message = e.get_HTTP()
```
xadahiya: sketch 43

Mec-iS:
@xadahiya I think we can make a custom exception class with a proper method and use that instead of `Exception`. You should open a related issue about this.

@chrizandr please provide indication to solve the error

vddesai1998:
I can make it work with python 3.5.2  and other python 3 versions by making few changes but we will still get those attribute errors.

vddesai1998:
In hydrus/exception.py we have some custom exceptions defined. How about using them?

hydrus/app.py
vddesai1998: Adapted to work with all python 3 versions

vddesai1998:
@xadahiya @Mec-iS  I have made required changes.
I think we can merge this PR now.

Meanwhile, I can open another PR to solve Exception handling problem this way.
** From this**
```
def get(self, id_, type_):
        """GET object with id = id_ from the database."""
        if get_authentication():
            if request.authorization is None:
                return failed_authentication()
            else:
                try:
                    auth = check_authorization(request, get_session())
                    if auth is False:
                        return failed_authentication()
                except Exception as e:
                    status_code, message = e.get_HTTP()
                    return set_response_headers(jsonify(message), status_code=status_code)

        class_type = get_doc().collections[type_]["collection"].class_.title

        if checkClassOp(class_type, "GET"):

            try:
                response = crud.get(id_, class_type, api_name=get_api_name(), session=get_session())
                return set_response_headers(jsonify(hydrafy(response)))

            except Exception as e:
                status_code, message = e.get_HTTP()
                return set_response_headers(jsonify(message), status_code=status_code)

        abort(405)
```

**To this**

```
    def get(self, id_: int, type_: str) -> Response:
        """GET object with id = id_ from the database."""
        if get_authentication():
            if request.authorization is None:
                return failed_authentication()
            else:
                try:
                    auth = check_authorization(request, get_session())
                    if auth is False:
                        return failed_authentication()
                except UserNotFound as e:
                    status_code, message = e.get_HTTP()
                    return set_response_headers(jsonify(message), status_code=status_code)

        class_type = get_doc().collections[type_]["collection"].class_.title

        if checkClassOp(class_type, "GET"):

            try:
                response = crud.get(id_, class_type, api_name=get_api_name(), session=get_session())
                return set_response_headers(jsonify(hydrafy(response)))

            except ClassNotFound as e:
                status_code, message = e.get_HTTP() # type: ignore
                return set_response_headers(jsonify(message), status_code=status_code)

            except InstanceNotFound as e:
                status_code, message = e.get_HTTP() # type: ignore
                return set_response_headers(jsonify(message), status_code=status_code)

        abort(405)
```

Yes, I have imported those custom Exception classes from hydrus/data/exceptions.py

hydrus/app.py
vddesai1998: This is an issue with Exception handling. Type annotations are fine. As you can see in the screenshot there are no type errors.
I will open one more PR to solve that exception handling issue.
Thanks for reviewing!

vddesai1998:
mypy

PR #108, opened 2018-02-20
Removed pdb import from data/user.py

py-ranoid:


chrizandr:
Please create one PR with all commits.

Issue #99, opened 2018-02-18
Lacks support for sqlalchemy.orm.scoped_session.


KaranSharma-gawds:
### I'm submitting a 
- [ ] bug report.

### Current Behaviour:

Lacks support for sqlalchemy.orm.scoped_session.
Currently the code in set_session() checks only for ```sqlalchemy.orm.session.Session```, while flask also supports making of other types of sessions.

### Expected Behaviour:

```scoped_session``` in sqlalchemy.orm will be supported

### Steps to repoduce:

Import scoped_session.
```python
from sqlalchemy.orm import scoped_session.
```
Replace the current session-creation code with the following:
```python
session = scoped_session(sessionmaker(autocommit=False, autoflush=False, bind=engine))
```
### Do you want to work on this issue ?

Yes.

xadahiya:
I think using `scoped_session` instead of `session` is a good idea.
Those who are not familiar with `scoped_session`, please read [this](http://docs.sqlalchemy.org/en/latest/orm/contextual.html).

@Mec-iS  @chrizandr  what do you think?

KaranSharma-gawds:
I was thinking of reducing the dependency on the kind of session object being used altogether rather than just switching it to another, as the issue of compatibility will still remain even if we change it from one to another. If some API is not using the session object included in here, then they will have to face this issue otherwise.

Mec-iS:
So in practice we should use `scoped_session` in our code and check for the existence of different classes of sessions? Using `scoped_session` makes sense as we may want the session to have the same lifetime as the request.

xadahiya:
Exactly!

alohamora:
Can I work on this issue?
## My Implementation
Check for both types of session in `utils.py`:
```python
if not isinstance(DB_SESSION,scoped_session) and not isinstance(DB_SESSION,Session):
```
Use scoped_session for thread-managed registry of database session in `main.py`:
```python
session = scoped_session(sessionmaker(autocommit=False, autoflush=False, bind=engine))

Mec-iS:
If you have a fix that pass all tests, you can open a PR and ask for a review.

Read the guidelines here: https://github.com/HTTP-APIs/hydrus/wiki/Starting-Material

xadahiya:
Fixed in #112

KaranSharma-gawds:
@Mec-iS @xadahiya please review the PR #162 .

PR #112, opened 2018-02-21
Added support for scoped_session

alohamora:
Issue #99 

Fixes #

### Checklist
- [ ] My branch is up-to-date with upstream/develop branch.

### Description
Added support for inititalising the database engine using `scoped_session` in `main.py` and in `get_session` function in `utils.py`  and also included instance check for both `Session` and `scoped_session` type

P.S. - This is my first PR

### DIFF
--- a/hydrus/utils.py
+++ b/hydrus/utils.py
@@ -5,7 +5,7 @@ from flask import appcontext_pushed
 from flask import g
 from hydrus.hydraspec import doc_writer_sample
 from hydrus.data.db_models import engine
-from sqlalchemy.orm import sessionmaker
+from sqlalchemy.orm import sessionmaker,scoped_session
 from sqlalchemy.orm.session import Session
 from hydrus.hydraspec.doc_writer import HydraDoc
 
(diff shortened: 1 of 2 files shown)


chrizandr:
You may also have to update the tests since they use session instead of scoped_session

alohamora:
I checked with the test module....there is no test asserting the type of session.
Adding to that...the `test_app.py` and `test_crud.py` module were passed successfully after the changes, but the `test_auth.py` module was giving `401 status code` assertion error even before any changes were applied.

chrizandr:
Since we are now shifting to scoped sessions instead of the normal session, it would make sense to use the same in tests. The tests in `test_app` and `test_crud` use session objects instead of scoped_session to run tests. See lime 44 in `test_app` for example.

alohamora:
Added `scoped_session` based database sessions in all the test files...Sorry for missing that in the first place.

alohamora:
@Mec-iS @chrizandr....completed all the tasks.
Please review the pull request

Issue #79, opened 2018-02-13
Coding Conventions


yashLadha:
Currently the code has many functions that doesn't follow proper naming conventions and also lines are greater than 79 characters, which is not according to Conventions and standards followed. 

Some of the functions with improper naming convention : 

![image](https://user-images.githubusercontent.com/18033231/36139099-3ddc9a16-10c2-11e8-8a48-888a8b69dc70.png)

__Working status__:
Yes I am working on it.

yashLadha:
Ping @chrizandr @Mec-iS

starlord1311:
@yashLadha  can i work on this?

yashLadha:
@starlord1311 I have clearly mentioned in the issue description that i am working on this. I am waiting for the response @chrizandr @Mec-iS

Mec-iS:
NOTE: I would prefer to stick to PEP-8. In particular functions' names should be lowercase and underscored (not camel-case). You can use `pylint` as checker. 

Furthermore, also implementing autodocumentation with Sphinx should be considered.

yashLadha:
Sure, I will try to do proper convention and for Sphinx i have to learn it and will update you on it @Mec-iS  😄

yashLadha:
@Mec-iS Should I also change the docstrings , as they are stretching to length > 79. This way i will get a better understanding of codebase as well. Views?

Mec-iS:
Not a great issue. I think we can leave the lines length as it is now. See [here](http://hilton.org.uk/blog/source-code-line-length):

> 125 characters per line is the real de facto coding standard for maximum line length these days, because this is the maximum number of characters that you can see in the GitHub diff view. This used to be 119 characters, but the page layout changed.

Usually 89 or 99 is also accepted. 79 could be too short for current screen sizes.

yashLadha:
But a you mentioned, in the PEP-8 conventions it is of 79 length and pylint follows the same. So should i lint the code according to PEP-8 conventions or leave it as it is and work on the renaming of functions.

chrizandr:
The values in pylint for number of characters can be modified. It is not a strict convention. Most codebases nowadays require more than 79 characters. There are also some lines that cannot be shortened. It's better to leave it as it is.

yashLadha:
Sure @chrizandr . Will make the updates.

IamRaviTejaG:
[...]

PR #116, opened 2018-02-23
Refactoring Key Validation Based on dictionary #95

### LINKED ISSUE #95: Refactoring key validation in a dictionary body
### I'm submitting a 
- [x] Enhancement Request.

### Current Behaviour:

We are using try catch to check every key in the dictionary and assigning it to new variable.
rather to validate the current payload we should have a generic function in which we can pass the key and value to check and it will return the keys or exception from that function 
[Current Sample code structure] (https://ibb.co/ciM4Vn)

### Do you want to work on this issue ?

yes
@Mec-iS -can you provide any suggestion here

anubhav253:
Fixes #95 

### Checklist


alohamora:
@anubhav253 I think that the `errorMapping` function is redundant as all the error strings have a similar format, so we can simply print the error string as:
`raise SyntaxError("The " + bodyType + " must have [" + key + "]")`
and pass the respective string in bodyType.

anubhav253:
@alohamora Yeah but I don't wanted to hamper the existing output as my target to refractor the code repetation and using dictionary here also to change the error message we can have a discussion with Mec-iS as I haven't had a talk regarding this.

Mec-iS:
Some basic things, wait for review: 
1. the code should be formatted as specified by PEP-8, check #79 
2. the standard way of formatting string is by using `str.format()` [doc](https://docs.python.org/3.5/library/stdtypes.html#str.format), check also [syntax](https://docs.python.org/3.4/library/string.html#format-string-syntax)

hydrus/hydraspec/doc_maker.py
```diff
@@ -16,56 +41,46 @@ def createDoc(doc: Dict[str, Any], HYDRUS_SERVER_URL: str=None, API_NAME: str=No
         raise SyntaxError("The API Documentation must have [@id]")
 
     # Extract base_url, entrypoint and API name
-    matchObj = re.match(r'(.*)://(.*)/(.*)/vocab#?', id_, re.M | re.I)
-    if matchObj:
-        base_url = matchObj.group(1) + '://' + matchObj.group(2) + '/'
-        entrypoint = matchObj.group(3)
+    match_obj = re.match(r'(.*)://(.*)/(.*)/vocab#?', id_, re.M | re.I)
+    if match_obj:
+        base_url = match_obj.group(1) + '://' + match_obj.group(2) + '/'
```
Mec-iS: String concatenation with `+` should be a rare occurrence. Wherever possible use string formatting `str.format` (rarely concatenation is preferred for performance reason).

anubhav253: @Mec-iS Its updated as per your suggestion.

anubhav253:
@xadahiya Type annotations are added and also updated the branch with develop.

anubhav253:
@xadahiya I have done the changes requested here please review it once.

xadahiya:
@chrizandr  ^^

PR #121, opened 2018-02-24
API authentication logic refactored

### LINKED ISSUE #120: Repetition of code
### I'm submitting a 
- [x] Code optimization.

### Current Behaviour:

This Logic to check whether a http request made is authenticated or not is repeated inside every CRUD operations.This whole logic is repeated at 8 places. Few samples are provided here.

https://github.com/HTTP-APIs/hydrus/blob/f88d0d06933ec0d426bd7d6df51f2972fd389a40/hydrus/app.py#L103

https://github.com/HTTP-APIs/hydrus/blob/f88d0d06933ec0d426bd7d6df51f2972fd389a40/hydrus/app.py#L131

https://github.com/HTTP-APIs/hydrus/blob/f88d0d06933ec0d426bd7d6df51f2972fd389a40/hydrus/app.py#L169

### Expected Behaviour:

Their should be a single function to check the authentication called by each CRUD operation.

yes

anubhav253:
Fixes #120 

### Checklist

### DIFF
--- a/hydrus/app.py
+++ b/hydrus/app.py
@@ -27,6 +27,21 @@ def failed_authentication() -> Response:
                                     headers=[{'WWW-Authenticate': 'Basic realm="Login Required"'}])
     return response
 
+def check_authentication() -> Response:
+    """Check the authentication for API CRUD operations.
+    :return return_type: response object
+    """
+    if get_authentication():
+        if request.authorization is None:
+            return failed_authentication()
+        else:
+            try:
+                auth = check_authorization(request, get_session())
+                if auth is False:
+                    return failed_authentication()
+            except Exception as e:
+                status_code, message = e.get_HTTP()  # type: ignore
+                return set_response_headers(jsonify(message), status_code=status_code)
 
 def set_response_headers(resp: Response, ct: str="application/ld+json", headers: List[Dict[str, Any]]=[], status_code: int=200) -> Response:
     """Set the response headers."""
@@ -107,17 +122,7 @@ class Item(Resource):
 
     def get(self, id_: int, type_: str) -> Response:
         """GET object with id = id_ from the database."""
-        if get_authentication():
-            if request.authorization is None:
-                return failed_authentication()
-            else:
-                try:
-                    auth = check_authorization(request, get_session())
-                    if auth is False:
-                        return failed_authentication()
-                except Exception as e:
-                    status_code, message = e.get_HTTP()  # type: ignore
-                    return set_response_headers(jsonify(message), status_code=status_code)
+        check_authentication()
 
         class_type = get_doc().collections[type_]["collection"].class_.title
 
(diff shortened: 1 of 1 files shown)


yashLadha:
Please check the necessary fields that are in your PR.

anubhav253:
@yashLadha  yeah I have updated and thanks for pointing it out.

hydrus/app.py
```diff
@@ -26,6 +26,18 @@ def failed_authentication():
                                     headers=[{'WWW-Authenticate': 'Basic realm="Login Required"'}])
     return response
 
+def check_authentication():
```
vddesai1998: LGTM.
But you should type annotate the new function you defined.

hydrus/app.py
```diff
@@ -27,6 +27,20 @@ def failed_authentication() -> Response:
                                     headers=[{'WWW-Authenticate': 'Basic realm="Login Required"'}])
     return response
 
+def check_authentication():
```
vddesai1998: You have to use python3 type annotation here.
Specify argument types and return type.
As your function is not taking any arguments you have to just specify the return type.

vddesai1998:
Why are you removing type annotations from the file?

xadahiya:
@anubhav253  Please resolve file conflicts.

anubhav253:
@vddesai1998  - please check the type annotations are already removed by someone else(in develop branch) i just have merged the changes of develop branch to my branch.

@xadahiya - can you please suggest what should i do here. Keep changes from the develop branch as it doesn't contain vddesai changes here.

Mec-iS:
@anubhav253 wait for the tyoe annotations to be all present in `develop` and then merge

chrizandr:
Have you actually run the server after making the changes?

anubhav253:
Yeah @chrizandr i had also started server and tried all the test cases here providing you the screenshots here.
screenshot 100

xadahiya:
Try running the actual server, it never asks for any authentication. Tests are passing because they are incomplete. See https://github.com/HTTP-APIs/hydrus/issues/140

anubhav253:
@xadahiya If I am expected here to run python main.py and make a curl request or I have to do something else here ? Please suggest !

screenshot 103

xadahiya:
Run the server and try using it simply in the browser. It won't ask authentication at any endpoint

chrizandr:
That's part of the reason why I asked wether you ran the server or not. The request object is context dependent on the function from which it is accessed. You can't access request outside of the get, put, post and delete functions. The same is true for the authorization flag, since it is always accessed outside the app context, the `check_authorization` function would return None, and there would be no checking done. Like I mentioned in the issue, we need to keep that part of the code within the get, put, post and delete functions or find a workaround for assigning variables that do not depend on function contexts.

anubhav253:
@chrizandr - ok so  the new function that i made will not have the application context here but i am getting similar behavior with our without these changes when i am restarting my server and trying to hit the api from browser its not asking for any authorization in(develop as well as this branch) thats why i am confused. Can you please help me out here to reproduce difference in those  so that i can debug and understand them more properly.

chrizandr:
If that is the case then we might need to see if any changes have been made that have caused authentication to not work. I will check the code today or tomorrow and let you know about that. For now, I am marking this PR as a work in progress.

anubhav253:
@chrizandr sure and thanks 🙂 will wait for your revert.

xadahiya:
@anubhav253  authentication is working on the old server. Try visiting `serverapi/DroneCollection` it'll ask for authentication.

anubhav253:
Thanks @chrizandr @xadahiya now i got what was missing here. i will try to find if it can be done by any means, will revert here if i could find nything by tomorrow.

Mec-iS:
This converged into #123

Issue #123, opened 2018-02-24
Generating user nonces for better security


IamRaviTejaG:
### I'm submitting a 
- [ ] bug report.
- [x] feature request.

### Current Behaviour:
Currently, the feature is still in TODO.

https://github.com/HTTP-APIs/hydrus/blob/f88d0d06933ec0d426bd7d6df51f2972fd389a40/hydrus/data/user.py#L22

### Expected Behaviour:
We can add a unique nonce by using random numbers generated using a seed, which can be the sha224 digest of user id or any other parameter.

### Do you want to work on this issue ?
Yes

@Mec-iS Please elaborate on this TODO.

PS: Working on this. PR coming up in a while.

Mec-iS:
duplicate #124

Mec-iS:
As specified in PR:

> I think this needs to be part of a mechanism of 2-factors authentication.
When client and server handshake, the server return an header with a nonce. Then the client authenticates with username, password and nonce. Once the client in authenticated, the server returns a (expiring) token so the client can authenticate with that, without the need for username and password.

> This is not an easy issue, take your time.

alohamora:
@Mec-iS The 2 factor authentication shall be used with `Basic authentication`  or `Digest authentication`??

alohamora:
And how should we send the token to the authentic user??

Mec-iS:
The simplest procedure we can implement is:
* client makes a request
* server responds with a randomly generated string (nonce)
* client passes username password and nonce 
* server checks credentials and nonce and responds with a token that expires in X days
* the client can connect to the endpoints using the token. When the token expires, the procedure is repeated

This way the process of Authentication and the process of Authorization are segregated and there is less password sharing because authorization is token-based.

alohamora:
@Mec-iS For now, I have made a PR #144 which supports the following implementation:

- client makes a request
- server responds with basic authentication challenge (wihout a nonce)
- client passes username and password 
- server checks credentials and responds with a token (valid for infinite days)
- the client can connect to the endpoints using the token.

IamRaviTejaG:
@Mec-iS How about using PGP? The client generates a key pair which is stored locally. The server will have it's own key pair. The client makes a request intended to the server which contains the data and once validated on the server, the server responds with a token. This reduces the load on server (no need to generate the nonce), and furthermore, any communication between the client and the server can also be secured via PGP.

alohamora:
[...]

PR #124, opened 2018-02-24
User nonces for better security

IamRaviTejaG:
Fixes #123

### Checklist

### Description
Generates a random number (user nonce) using a random seed which is the sha224 digest of the user id.

### DIFF
--- a/hydrus/data/user.py
+++ b/hydrus/data/user.py
@@ -6,7 +6,7 @@ from hydrus.data.exceptions import UserExists, UserNotFound
 from hydrus.data.db_models import User
 from hashlib import sha224
 import base64
-# import random
+import random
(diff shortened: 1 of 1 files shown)


IamRaviTejaG:
@Mec-iS @xadahiya The feature was on the TODO. Please elaborate on how and where to use it.

Mec-iS:
I think this needs to be part of a mechanism of 2-factors authentication. 
When client and server handshake, the server return an header with a nonce. Then the client authenticates with username, password and nonce. Once the client in authenticated, the server returns a (expiring) token so the client can authenticate with that, without the need for username and password.

Mec-iS:
this converge into #144

Issue #128, opened 2018-02-25
Missing mypy type annotations


xadahiya:
### I'm submitting a 
- [X] bug report.

### Current Behaviour:

Mypy type annotations in some files are gone after merging #117 
### Expected Behaviour:

All files should have mypy type annotations
### Steps to reproduce:

See app.py and several other files

PR #131, opened 2018-02-25
revamp tests

### LINKED ISSUE #129: Make Hydrus utilities available for entire test class rather than specific test function.
### I'm submitting a 
- [x] bug report.
- [ ] feature request.

### Current Behaviour:

As of now, for each test function, the `hydrus`  utilities are set up using the nested `with` statements.
```python
 with set_api_name(self.app, self.API_NAME):
            with set_session(self.app, self.session):
                with set_doc(self.app, self.doc):
                    with self.app.test_client() as client:
```
This is apparently not a good approach because the setting up and destruction of contexts take place for every test function.

### Expected Behaviour:

A better, alternative approach will be to set up contexts once in `setUpClass`, use them throughout the tests and then destroy them in `tearDownClass`. This will allow new developers to focus on writing tests rather than setting up contexts. It will also improve the test run time since setting up and destruction process takes place only once.

Yes.
I am looking at ways to achieve this. Will send a PR once done.

mohdomama:
Fixes #129  

### Checklist

### Description

The `hydrus` utilities were being set up and destroyed for each test function. To prevent this, I am setting up utilities in the `setUpClass`, using them for all the tests and then destroying them in the `tearDownClass`. To achieve this, I have used the implicit  `__enter__()` and `__exit__()` methods of the `contextmanager` decorator.

### Test Logs

hydrus

### DIFF
--- a/hydrus/tests/test_app.py
+++ b/hydrus/tests/test_app.py
@@ -56,315 +56,269 @@ class ViewsTestCase(unittest.TestCase):
         print("Classes and properties added successfully.")
         print("Setup done, running tests...")
 
+        self.api_name_util =  set_api_name(self.app, self.API_NAME)
+        self.session_util = set_session(self.app, self.session)
+        self.doc_util = set_doc(self.app, self.doc)
+        self.client = self.app.test_client()
+        
+        self.api_name_util.__enter__()
+        self.session_util.__enter__()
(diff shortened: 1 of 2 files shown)


xadahiya:
@chrizandr  @Mec-iS  please don't merge unless #128 is fixed.

Issue #140, opened 2018-02-27
Incomplete authentication test suite


xadahiya:
### I'm submitting a 
- [X] bug report.
- [X] feature request.

### Current Behaviour:

Authentication tests need to be updated. Current tests only test `/+ API_NAME`, rest everything needs to be tested as well.
### Expected Behaviour:

Tests should check the authentication part thoroughly.

### Steps to reproduce:

See `hydrus/hydrus/tests/test_auth.py`.
### Do you want to work on this issue?

NO

vaibhavchellani:
@xadahiya i would like to work on this . I will try to submit a PR asap . Any points/tips you would like to offer ?

xadahiya:
@vaibhavchellani  Sure! First, try using the Hydrus server in your browser. There are certain endpoints     (**Collection endpoints**) that need users to log in before they can be accessed.

See this
![sketch 45](https://user-images.githubusercontent.com/9114274/36748665-8debf210-1c1e-11e8-9f08-7a750fe2f7e8.png)

You should try to test different behaviors like what happens when some unauthorized user visits a collection endpoint. 
Think of different scenarios for various types of endpoints and write tests.

tibi77:
Maybe if we do not show the endpoints. We can create a specific file (for example http://localhost:8080/serverapi/endPoints) which can be password protected. I think that it will be safer.

Right now: If someone who is not authorized tries to login the server responds with:
```
{
  "message": "The User with ID <user_id> is not a valid/defined User"
} 
```
I have found this checklist[1]. Should i implement the checks that are there?
Also the pentest part in [2] is interesting.
PS: I am asking because i am not sure what behavior you are expecting.

[1] https://www.owasp.org/index.php/REST_Security_Cheat_Sheet#Validate_content_types
[2] https://www.owasp.org/index.php/REST_Assessment_Cheat_Sheet

chirag-jn:
I have added test cases for invalid id and invalid password by changing the id and password in the test cases from default "1" and "test" respectively.

xadahiya:
We still need tests to check how hydrus responds when unauthorized users visit different server endpoints like collection endpoints (For example `/serverapi/DroneCollection/1`).

tibi77:
Right now it hydrus is associating my ip(or other token of some sorts) with my ip and pass. Wouldn't it be better to ask for authentication every time we try to connect?
Also I think i can do the different users endpoints thing :smiley:.

[...]

PR #144, opened 2018-03-01
Added token based authentication feature for user

alohamora:
Fixes #123 #120 

### Checklist

### Description

These changes allow an authenticated user to get a token after authenticating with the server and the token allows the user to access the protected resources of the API by using the token in the url as:
`http:localhost:8080/serverapi/SomeCollection?token="API_TOKEN"` without authorizing itself.
Also this fixes the repetition of code in `app.py` module by combining the authentication steps in on e function #120 (tested the server on browser)

### Test Logs

tests
### Sample Usage Explanation
1. The user authenticates itself with the server using username and password.

2. After succesful authentication, the server responds with a server token for further use

3. This token can then be used to directly acces the protected endpoints from any source.

NOTE:
1. For now the server issued tokens are generated for infinite time, will add time based destruction of the token in later stages.
2. I have added a `set_token` contextmanager function for now to enable or disable this feature in the API
3. For `auth_tests` currently the `set_token` feature is set to False to comply with the original tests.

### DIFF
--- a/hydrus/app.py
+++ b/hydrus/app.py
@@ -82,6 +82,7 @@ def checkClassOp(class_type: str, method: str) -> bool:
     return False
 
 def verify_user() -> Union[Response, None]:
+    """ Verify the credentials of the user and assign token."""
     try:
         auth = check_authorization(request, get_session())
         if auth is False:
--- a/hydrus/app.py
+++ b/hydrus/app.py
@@ -21,7 +21,7 @@ def validObject(object_: Dict[str, Any]) -> bool:
 
 def token_response(token: str) -> Response:
     """Return succesful token generation object"""
-    message = {"User token": token}
+    message = {200: "User token generated"}
     response = set_response_headers(jsonify(message), status_code=200,
                                     headers=[{'X-Authorization': 'TOKEN ' + token}])
     return response
@@ -36,7 +36,7 @@ def failed_authentication(incorrect: bool) -> Response:
         realm = 'Basic realm="Incorrect credentials"'        
     nonce = create_nonce(get_session())
     response = set_response_headers(jsonify(message), status_code=401,
-                                    headers=[{'WWW-Authenticate': realm},{'Set-Cookie': 'nonce=%s' %nonce}])
+                                    headers=[{'WWW-Authenticate': realm},{'X-Authentication': 'NONCE %s' %nonce}])
     return response
 
 
@@ -106,7 +106,7 @@ def verify_user() -> Union[Response, None]:
 def check_authentication_response() -> Union[Response,None]:
     """ Returns the response as per the authentication requirements."""
     if get_authentication():
-        if request.args and get_token():
+        if get_token():
             token = check_token(request, get_session())
             if not token:
                 if request.authorization is None:
--- a/hydrus/app.py
+++ b/hydrus/app.py
@@ -23,7 +23,7 @@ def token_response(token: str) -> Response:
     """Return succesful token generation object"""
     message = {200: "User token generated"}
     response = set_response_headers(jsonify(message), status_code=200,
-                                    headers=[{'X-Authorization': 'TOKEN ' + token}])
+                                    headers=[{'X-Authorization': token}])
     return response
 
 def failed_authentication(incorrect: bool) -> Response:
@@ -36,7 +36,7 @@ def failed_authentication(incorrect: bool) -> Response:
         realm = 'Basic realm="Incorrect credentials"'        
     nonce = create_nonce(get_session())
     response = set_response_headers(jsonify(message), status_code=401,
-                                    headers=[{'WWW-Authenticate': realm},{'X-Authentication': 'NONCE %s' %nonce}])
+                                    headers=[{'WWW-Authenticate': realm},{'X-Authentication': nonce}])
     return response
 
 
--- a/hydrus/app.py
+++ b/hydrus/app.py
@@ -81,30 +81,34 @@ def checkClassOp(class_type: str, method: str) -> bool:
             return True
     return False
 
+def verify_user() -> Union[Response, None]:
+    try:
+        auth = check_authorization(request, get_session())
+        if auth is False:
+            return failed_authentication()
+        else:
+            if get_token():
+                token = add_token(request, get_session())
+                return token_response(token)
+    except Exception as e:
+        status_code, message = e.get_HTTP()  # type: ignore
+        return set_response_headers(jsonify(message), status_code=status_code)
+    return None  
+
 def check_authentication_response() -> Union[Response,None]:
     """ Returns the response as per the authentication requirements."""
     if get_authentication():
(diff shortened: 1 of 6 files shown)


alohamora:
Added a basic token based 2 factor authentication.
@xadahiya @Mec-iS @chrizandr ...please review the request.

xadahiya:
sketch 47
The server always returns user token, no matter what URL I visit. It should return data not user token

alohamora:
@xadahiya I also face that error...its because firefox send the authentication headers everytime after the first authentication request is sent.
For now my logic returns the token to the user everytime the auth header is passed in the URL.
To test the token, you can open the URL with token using curl or in a private window....it shall work.

alohamora:
@xadahiya fixed repetition of token response...please check for the same.

alohamora:
git2
@xadahiya fixed the response of the page....please review

alohamora:
@Mec-iS please suggest the changes to be added in the PR.

Mec-iS:
I think my opinion is the same I specified in #123

alohamora:
Actually this PR included only a basic token based authentication without expiration. Also for the nonce based authentication I can't find a way to authenticate the user using the three credentials(username,password and nonce) except for using `digest based authentication`.
So, can you please guide me on that.

alohamora:
Added the expiration feature in tokens. For now the time to expire for the tokens is 1min, please suggest me the apt time of expiration to set for the tokens. @Mec-iS @xadahiya

alohamora:
@Mec-iS  @xadahiya  Added nonces for security, here is the final stepwise description of the authentication system:
- The client requests a protected resource.
- The server responds with a `401` response along with a nonce passed in the `Set-Cookie` header.
- The client authenticates with username, password and the nonce.
- The server verifies the nonce, deletes it and then returns the token to the user.
- The client can then access the protected resource using the token only.

NOTE:
- If using a web browser, the nonce will be passed by the browser automatically as we've used `Set-Cookie`.
- If using curl request, the nonce has to be passed in the cookie header as `Cookie: nonce="nonce-value"`
- Both tokens and nonces are time bound, currently nonces are valid for 1 min and tokens for 2 min.
- This change shall lead to failure of auth tests for now.

alohamora:
Here is a brief demo of the auth system:
1. The user requests protected resource, gets a 401 and nonce in response

first-req

2. The user requests the resource again, with the nonce cookie and the auth header, gets a token in return

second-req

3. The user uses the token to access the resource directly

third-req

xadahiya:
Nonce expiry time is fine but I think for the token it should be around 30 mins to 1 hour. 
@Mec-iS ^^

Mec-iS:
@alohamora use `X-Authentication` as header name when passing the nonce, and `X-Authorization` when passing the token.

alohamora:
@Mec-iS  Sir, how shall the client send the nonce back to the server while using  `X-Authentication`, currently I am using the cookie for automatic cookie response from the browser.
Also do I remove the token from the response data or let it be there??

Mec-iS:
@alohamora If you want to make the authentication works on the browser for visualization/debugging purpose you can use the cookie work-around, but HYDRA is meant to work for automated clients that usually use lower-level clients and not browsers. So the procedure shall work first using `curl` or `requests`, they both can set arbitrary headers to arbitrary values. Then it can also work in the browser if the server is in debugging mode.

Token should be sent via header by the client per-request and checked by the server. The server shall not write the token back as it is superfluous, and anything superfluous usually hits efficiency.

alohamora:
@Mec-iS  Here is the summary of changes done as per the specifications:
- The server sends the nonce as `X-Authentication: NONCE "nonce-value"`
- The client has to send the nonce to the server using `nonce:"nonce-value"` header.
- The server sends the token generated as `X-Authorization: TOKEN "token-value"`
- The client has to send the token to the server using `token: "token-value"` header.
- The token expiry time is set to `45 min`
- The server sends a `401` response along with `incorrect credentials` message if the nonce or credentials are incorrect or expired.

Mec-iS:
> The server sends the nonce as X-Authentication: NONCE "nonce-value"
The client has to send the nonce to the server using nonce:"nonce-value" header.
The server sends the token generated as X-Authorization: TOKEN "token-value"
The client has to send the token to the server using token: "token-value" header.

Semantics for client and server should be the same. Differentiating syntax from client and server makes only the protocol more difficult to understand in my opinion:
* The server sends the nonce as `X-Authentication`: NONCE "nonce-value"
* The client has to send the nonce to the server using `X-Authentication`: NONCE "nonce-value" along with username and password (in the body).
* The server sends the token generated as `X-Authorization`: TOKEN "token-value"
* The client has to send the token to the server using `X-Authorization`: TOKEN "token-value" in every request to a protected endpoint

Please write extensive documentation and PR the Wiki with a file called `Auth.md` at [here](https://github.com/HTTP-APIs/hydra-ecosystem-wiki)

alohamora:
> The client has to send the nonce to the server using X-Authentication: NONCE "nonce-value" along with username and password (in the body).

Sir, currently we authenticate the credentials using `basic authentication header`, so do you mean to say that the client shall pass its credentials as a json in request data `{username:"1", password:"test"}` instead of using the authentication header?
And, correspondingly should there be any changes in `WWW-Authenticate` header while sending the failed authentication response??

xadahiya:
@alohamora please sync with develop.

alohamora:
- matched user semantics for nonce and token response as `X-Authentication:nonce-value` and `X-Authorization:token-value` respectively for simplicity.
- synced with develop and updated documentation for functions.
@xadahiya @Mec-iS  please review the changes.

chrizandr:
@alohamora The use of the Basic protocol for authentication was a design choice used to get things working in the initial stages, the use of headers is the best way for low level clients to authenticate. It's not really useful to have tokens that you pass in the URL, because GET reqeusts over HTTP can be easily captured and decoded. If we can implement something that is more secure than HTTP Basic Authorization while still allowing low-level clients to use it, we can go ahead with it.

alohamora:
@chrizandr I have shifted the passing of auth tokens and nonces to the `X-Authentication` and `X-Authorization` header respectively.
Also I was thinking to shift the credentials to user data just because the low level clients like curl had to specifically encode the `username:password` string to pass in the header.
But I understand the concern and I shall search for a better alternative to this system as soon as possible.

alohamora:
@chrizandr Also please comment if any changes are needed to be done in this PR

alohamora:
Sir the server is running fine....here are the server logs and request logs:
requestlogs

serverlog

For the tests, the `test_app` and `test_crud` module is running fine, but the `test_auth` gives a failure because the tests are still not designed to send requests with nonces.
So, shall I mention this failure of tests in #140 or should I add commits in this PR only??

alohamora:
@chrizandr changed tests as per the new authentication architecture, also I have added tests to check the response for passing a wrong nonce....please review
test_auth

PR #166, opened 2018-03-10
CLI code refactoring

### LINKED ISSUE #165: Differences in cli.py and main.py
### I'm submitting a 
- [ ] bug report.
- [x] feature request.

### Current Behaviour:
Currently, the files `cli.py` and `main.py` are independent and any changes made in the main.py file are not reflected back into the CLI interface. This way, the CLI runs an older version of the code.

### Expected Behaviour:
The CLI is expected to run always the latest version of the code, which will be possible only when the `main.py` and `cli.py` are somehow merged.

Yes

IamRaviTejaG:
Fixes #165 

### Checklist

### Description
CLI code is now refactored. The `main.py` is now located in the `hydrus` folder. Only the `cli.py` can now be tweaked a little and all the major changes can now be made to `main.py` without affecting the CLI. This eases the maintenance of the CLI.

### Test Logs
Not required.


IamRaviTejaG:
@Mec-iS @xadahiya Please review this PR. Made changes as suggested. Using this approach, the CLI code doesn't need to undergo changes everytime the `main.py` is changed.

Mec-iS:
I think in this case you can move all the code in `cli.py` and delete `main.py` as it is not holding any generic functionality, but it is only a script to run a demo.

cli.py
```diff
@@ -32,8 +32,73 @@
 @click.argument("serve", required=True)
 def start(adduser, api, auth, dburl, hydradoc, port, serverurl, token, serve):
     """Python Hydrus CLI"""
```
Mec-iS: This docstring is not informative. The parameters description is missing.
Every method should be documented as:
```
def test_method(arg1, arg2, ...):
    """
    Description of what the method performs.

    Generic Notes to consider when running the method if any

   : param arg1: what kind of value is expected
   : param arg2: what kind of value is expected
   :return : What is returned or `None`
"""
```

Also add a general documentation for CLI usage and features at https://github.com/HTTP-APIs/hydra-ecosystem-wiki/issues/10

cli.py
```diff
@@ -32,8 +32,73 @@
...
-    HydrusServer.server(adduser, api, auth, dburl, hydradoc,
-                        port, serverurl, token, serve)
+    # The database connection URL
+    # See http://docs.sqlalchemy.org/en/rel_1_0/core/engines.html#sqlalchemy.create_engine for more info
+    # DB_URL = 'sqlite:///database.db'
+    DB_URL = dburl
+
+    # Define the server URL, this is what will be displayed on the Doc
+    HYDRUS_SERVER_URL = serverurl + ":" + str(port) + "/"
```
Mec-iS: when possible, good practice is to use `str.format()`

chrizandr:
I think we should keep `main.py` at the end of the day, hydrus is a python library. `main.py` shows the usage of hydrus as a Python package. Maybe move and rename it to demo.py?

Mec-iS:
Having a `main.py` file is not mentioned in any official Python documentation. `setup.py` defines the entrypoint. Maybe is more appropriate to have it as a basic test?

vddesai1998:
I think we should keep main.py and load required code in cli.py from main.py.If in future someone want to create GUI for server setup he/she would be able to reuse this code from main.py.  And also as @Mec-iS suggested it can be used as basic test.

IamRaviTejaG:
@vddesai1998 Please have a look at this commit https://github.com/HTTP-APIs/hydrus/pull/166/commits/a4b4f37a645ad5ea66a3b1680f809b838bfe7adc

Is this what you are suggesting, or is  it something different?

vddesai1998:
Yes, that looks good to me.
Let maintainers decide whether we are keeping main.py or not then we can think about reusability.

chrizandr:
@Mec-iS Yeah, there is no mention of that in the documentation. What I meant was that we should keep some sort of example script to show the usage of the Python API for hydrus. We can name it something other than `main.py`, but I think that not having a demo script and shifting everything to the CLI would be a little confusing. We can keep the CLI and the demo as separate files. 

The demo script would include detailed comments and usage instructions. Those who do not use the CLI would need a demo script to be able to setup the server.

IamRaviTejaG:
@Mec-iS So, should I backtrack my changes to commit https://github.com/HTTP-APIs/hydrus/pull/166/commits/a4b4f37a645ad5ea66a3b1680f809b838bfe7adc and add detailed docstrings?

xadahiya:
I agree we should keep main.py . We can either rename it or add it to the legacy folder. It provides a pretty good understanding of how things work with hydrus.

IamRaviTejaG:
@xadahiya How about leaving it inside the `hydrus` folder and use it as an import for `cli.py`?

IamRaviTejaG:
@Mec-iS Please suggest what needs to be done here.

Mec-iS:
@IamRaviTejaG Chris gave the indication and I supported his comment

Mec-iS:
IMO this can be merged

PR #179, opened 2018-03-17
Added deletion of expired tokens

### LINKED ISSUE #178: Expired token should be removed from the database
### I'm submitting a 
- [x] bug report.

### Current Behavior:
We are storing the token in our database which get expire after 45mins. With this expiration the token should also be removed from the database. As it will unnecessarily increase the size of database.

https://github.com/HTTP-APIs/hydrus/blob/a13f3d0a5cc1d2e62bb0c5e2d67705f962b6ad5c/hydrus/data/user.py#L88

### Expected Behavior:
We should have removed the token once its expired and we get a call with this token(But this one is not a full proof  as sometimes it might not be possible that we don't get a call that token). 
So here i would like to use redis which is commonly used for storing access-token with some expiration time. Which will also improve the efficiency of authenticated api because of its being faster that SQL.

### Steps to reproduce:
We can check the database as the key are not being removed.

yes

alohamora:
Fixes #178 

### Checklist

### Description

Added deletion of expired tokens and nonces.
Added index to the `token.id` and `token.timestamp` column and respectively for nonce for quick search.
Fixed error message to `401:incorrect credentials`  from `401:need credentials to authenticate` when a wrong token is used 
### Test Logs

tests

### DIFF
--- a/hydrus/app.py
+++ b/hydrus/app.py
@@ -148,9 +148,9 @@ def check_authentication_response() -> Union[Response,None]:
     if get_authentication():
         if get_token():
             token = check_token(request, get_session())
-            if not token:
+            if not token['valid']:
                 if request.authorization is None:
-                    return failed_authentication(False)
+                    return failed_authentication(token['present'])
                 else:
                     return verify_user()
         elif request.authorization is None:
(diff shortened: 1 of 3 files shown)


alohamora:
@Mec-iS @xadahiya  please review

xadahiya:
Moved to #281

Issue #149, opened 2018-03-02
Open API to Hydra Parser


suryatejreddy:
### I'm submitting a 
- [ ] bug report.
- [x] feature request.

To implement this, we can use existing Hydrus classes to create objects from the Open API doc. The following workflow was recommended by @Mec-iS . 
OpenAPI YAML -> hydrus classes -> HYDRA ApiDoc
I am currently working on porting the basic HYDRA functionalities (GET, PUT, POST methods; classes, collections)

More suggestions are welcome.

### Do you want to work on this issue?

yes

Mec-iS:
NOTE: In the beginning it is possible to start with a Hydra ApiDoc >> OpenAPI parser if it is an easier way

suryatejreddy:
@Mec-iS could you elaborate on the Hydra ApiDoc >> OpenAPI parser. So you want to implement a parser that converts a Hydra ApiDoc to an OpenAPI.

Mec-iS:
Probably it is too difficult for now to code a parser OpenAPI->ApiDoc. Instead doing an ApiDoc->OpenAPI one may be doable.

suryatejreddy:
Okay. Any suggestions on how to start? @Mec-iS

Mec-iS:
Exactly.

`hydrus` is already a parser from JSON-LD (HydraClasses) to Python classes. From these Python classes it is needed to create an OpenAPI-compliant document (as JSON or YAML file).

suryatejreddy:
A get request on `/serverapi/dummyClassCollection` returns the following json

```json
{
  "@context": "/serverapi/contexts/dummyClassCollection.jsonld", 
  "@id": "/serverapi/dummyClassCollection/", 
  "@type": "dummyClassCollection", 
  "members": [
    {
      "@id": "/serverapi/dummyClassCollection/1", 
      "@type": "dummyClass"
    }
  ]
}
```
However, there is no way to access the `dummyClass` object. Because `/serverapi/dummyClassCollection/1` returns the following error. 

```json
{
"message": "The method is not allowed for the requested URL."
}
```
Shouldn't the first get method return a list of `dummyClassColelction` items with the respective properties? 
Or is there something i am missing out on? @Mec-iS 

I am trying to convert the `Collection` endpoints to OpenAPI path objects and then develop a web interface for hydrus where users can edit the HydraAPI and make changes just like swagger hub.

chrizandr:
[...]

PR #196, opened 2018-05-18
Hydra objects to OpenApi Objects.

### LINKED ISSUE #195: Basic Implementation Hydra ->Open Api parser
### I'm submitting a 
- [ ] bug report.
- [X] feature request.

Basic implementation of Hydra to Open Api definition , will converts most used objects of both and will generate a YAML file from Hydra Api Doc . 
Yes !

vaibhavchellani:
Fixes #195 

### Checklist
- [X ] My branch is up-to-date with upstream/develop branch.
- [X ] Everything works and tested for Python 3.5.2 and above.

### Test Logs

screen shot 2018-05-18 at 11 24 16 am

#### Added

- PyYaml to generate yaml file
- Basic structure of Open Api Definition 
- Can parse classes and make them definitions in OAS
- Can parse properties and descriptions and add as child objects of definitions

### DIFF
--- a/hydra_openapi_parser.py
+++ b/hydra_openapi_parser.py
@@ -1,0 +1,83 @@
+from hydrus.hydraspec.doc_maker import create_doc
+from hydrus.hydraspec.doc_writer_sample_output import doc
+
+import yaml
+from hydrus.data.doc_parse import get_classes, get_all_properties
+
+# below stuff is temporary
+apidoc = create_doc(doc, "http://localhost:8080/", "api")
+apiSpec = apidoc.generate()
+
+dataDump = dict(
+    {
+        "swagger": '2.0',
+        "info": {
+            "title": apiSpec["title"],
+            "description": apiSpec["description"]
+        },
+        "host": "hydrus.com",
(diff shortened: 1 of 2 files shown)


chrizandr:
I notice that you are trying to convert a Hydra API doc to OpenAPI, what we need however is for OpenAPI -> HydraAPI doc.
The reason we wanted a module for this is so that people who use OpenAPI, can easily use __hydrus__ without having to create separate API documentation, by directly using the parser.

vaibhavchellani:
yes i know the requirement is for Open Api -> Hydra , but following the discussion in #149 and the reasons i mentioned in my proposal i went for Hydra->Open Api . I thought i could make sure this way that we were able to map every object correctly , now that i have looked at both the specs closely enough i think we can directly go for Open Api -> Hydra , what do you suggest ?

xadahiya:
I think you should start working on openapi > hydra parser. If you get stuck at some point then we can figure it out along the way.

vaibhavchellani:
okay @xadahiya sounds cool !

PR #197, opened 2018-05-26
Open Api to Hydra Parser basic implementation 

### LINKED ISSUE #193: Mapping Hydra Objects to OpenAPI Objects.
Hydra representations of Objects needs to be mapped with OpenAPI objects.

The end goal is that given an OpenAPI Spec, we can generate an equivalent Hydra API Documentation that represents an API with the same functionality.

All discussions from #149 apply here.

Please specify features of the OpenAPI spec and we can discuss equivalent functionality in the Hydra spec.

vaibhavchellani:
Fixes #193 and #198 

### Checklist

### Description

This script is able to parse the open api spec and is able to extract the classes , supported props , supported operations from the file and create a Hydra Api Documentation . 

### Test Logs

@xadahiya @chrizandr @Shekharrajak , let me know the changes , its WIP


hydrus/hydraspec/openapi-hydra_parser.py
```diff
@@ -0,0 +1,113 @@
...
+    except yaml.YAMLError as exc:
+        print(exc)
+
+info = doc["info"]
+desc = info["description"]
+title = info["title"]
+baseURL = doc["host"]
+name = doc["basePath"]
+api_doc = HydraDoc(name, title, desc, name, baseURL)
+definitions = doc["definitions"]
```
xadahiya: Definitions won't always be defined in doc["definitions"]. You should use `$ref` see this https://swagger.io/docs/specification/using-ref/

hydrus/hydraspec/openapi-hydra_parser.py
```diff
@@ -0,0 +1,113 @@
...
+    for class_ in definitions:
+        try:
+            desc = definitions[class_]["description"]
+        except KeyError:
+            desc = class_
+        classDefinition = HydraClass(class_, class_, desc, endpoint=False)
+        properties = definitions[class_]["properties"]
+        classAndClassDefinition[class_] = classDefinition
+        definitionSet.add(class_)
+        api_doc.add_supported_class(classDefinition, collection=False)
```
xadahiya: How are you checking if any endpoint is a collection or not?

hydrus/hydraspec/openapi-hydra_parser.py
```diff
@@ -0,0 +1,113 @@
...
+                op_returns = None
+                op_status = [{"statusCode": 200, "description": "dummyClass updated"}]
+                try:
+                    op_name = paths[path][method]["summary"]
+                except KeyError:
+                    op_name = possiblePath
+                try:
+                    parameters = paths[path][method]["parameters"]
+                    for param in parameters:
+                        op_expects = param["schema"]["$ref"].split('/')[2]
```
xadahiya: Should be `'vocab:'+param["schema"]["$ref"].split('/')[2]`. ALso what if the parameter is something as simple as an integer and it's not defined in schema?

For example -  
``` yaml
      parameters:
        - name: petId
          in: path
          description: ID of pet to return
          required: true
          type: integer
          format: int64
```

hydrus/hydraspec/openapi-hydra_parser.py
```diff
@@ -0,0 +1,113 @@
...
+    we check if the path has a class defined , if it does we parse the methods and add the information parsed to Hydra
+
+    :return:
+    """
+    for path in paths:
+        possiblePath = path.split('/')[1]
+        # dirty hack , do  case insensitive search more gracefully
+        possiblePath = possiblePath.replace(possiblePath[0], possiblePath[0].upper())
+        # check if the path name exists in the classes defined
+        if possiblePath in definitionSet and len(path.split('/')) == 2:
```
xadahiya: What about paths like `/user/createWithList` or any other path with `length > 2`?

hydrus/hydraspec/openapi-hydra_parser.py
```diff
@@ -0,0 +1,113 @@
...
+                    for param in parameters:
+                        op_expects = param["schema"]["$ref"].split('/')[2]
+                except KeyError:
+                    op_expects = None
+                classAndClassDefinition[possiblePath].add_supported_op(HydraClassOp(op_name,
+                                                                                    op_method,
+                                                                                    "vocab:" + op_expects,
+                                                                                    op_returns,
+                                                                                    op_status))
+                api_doc.add_supported_class(classAndClassDefinition[possiblePath], collection=False)
```
xadahiya: Again, will only work for paths like `/User`. Also, what about collection type?

**Important - You're adding same class twice to the doc. First in line 41 and then again here.**
Why?

hydrus/hydraspec/openapi-hydra_parser.py
```diff
@@ -0,0 +1,113 @@
...
+desc = info["description"]
+title = info["title"]
+baseURL = doc["host"]
+name = doc["basePath"]
+api_doc = HydraDoc(name, title, desc, name, baseURL)
+definitions = doc["definitions"]
+paths = doc["paths"]
+definitionSet = set()
+classAndClassDefinition = dict()
+hydra_doc = ""
```
xadahiya: Please avoid using so many Global variables. You're not using function arguments **at all**. 
It's not a good practice.

hydrus/hydraspec/openapi-hydra_parser.py
```diff
@@ -0,0 +1,113 @@
+"""
+Module to take in Open Api Specification and convert it to HYDRA Api Doc
+
+"""
+import yaml
+from doc_writer import HydraDoc, HydraClass, HydraClassProp, HydraClassOp
```
xadahiya: Make sure you use absolute imports not relative imports.

hydrus/hydraspec/openapi-hydra_parser.py
vaibhavchellani: yes you are right , right now i am first parsing definitions and adding props and classes to api doc , then i parse paths and check if the path has the same name as class , if yes i add the operations .
Would the following be a better implementation , parse the paths , when $ref is found , parse whatever is defined and create classes ?

hydrus/hydraspec/openapi-hydra_parser.py
vaibhavchellani: I am not right now , in our discussions we decided to do that later .

hydrus/hydraspec/openapi-hydra_parser.py
vaibhavchellani: I have planned on adding some datatypes which Open api support like string , integer etc as classes and add those when we encounter them . I am adding the "vocab" part later in line 94

hydrus/hydraspec/openapi-hydra_parser.py
vaibhavchellani: yes , want to discuss about what to do about those .

hydrus/hydraspec/openapi-hydra_parser.py
vaibhavchellani: there would be some classes which are only made to be referenced later on , classes are added when their parsing is done , later if we find operations on them we add the operations and add it to the api doc again , the class appears only once in the final doc . Please suggest a better way .

hydrus/hydraspec/openapi-hydra_parser.py
vaibhavchellani: okay , thank you , i agree , will do .

hydrus/hydraspec/openapi-hydra_parser.py
vaibhavchellani: done !

hydrus/hydraspec/openapi-hydra_parser.py
xadahiya: Yeah, you should check where $ref points and then extract the definitions from there. It can be in the same doc or at some remote location but for now using local definitions should be fine

vaibhavchellani:
@xadahiya can this PR be merged ?

xadahiya:
@chrizandr  ^^

xadahiya:
@vaibhavchellani reduce the number of global variables a bit. Rest everything looks fine.

vaibhavchellani:
@xadahiya @chrizandr done .

hydrus/parser/openapi-hydra_parser.py
```diff
@@ -0,0 +1,121 @@
...
+from hydrus.hydraspec.doc_writer import HydraDoc, HydraClass, HydraClassProp, HydraClassOp
+
+with open("../samples/openapi_sample.yaml", 'r') as stream:
+    try:
+        doc = yaml.load(stream)
+    except yaml.YAMLError as exc:
+        print(exc)
+
+
+definitionSet = set()
```
chrizandr: Declare these in the `getClasses()` function and pass it as a variable to `get_ops()`. Whenever you can replace global variables with passable parameters, you should opt for the latter. `hydra_doc` can be declared in `getClasses()` as well and can be returned from to main when function finishes execution.

hydrus/parser/openapi-hydra_parser.py
```diff
@@ -0,0 +1,121 @@
+"""
+Module to take in Open Api Specification and convert it to HYDRA Api Doc
+
+"""
+import yaml
+import json
+
+from hydrus.hydraspec.doc_writer import HydraDoc, HydraClass, HydraClassProp, HydraClassOp
```
chrizandr: Good use of doc writer.

hydrus/parser/openapi-hydra_parser.py
```diff
@@ -0,0 +1,121 @@
...
+
+with open("../samples/openapi_sample.yaml", 'r') as stream:
```
chrizandr: Move this to main as well

hydrus/parser/openapi-hydra_parser.py
```diff
@@ -0,0 +1,121 @@
...
+                        op_expects = param["schema"]["$ref"].split('/')[2]
+                except KeyError:
+                    op_expects = None
+                try:
+                    responses = paths[path][method]["responses"]
+                    op_status = responses
+                except KeyError:
+                    op_returns = None
+                classAndClassDefinition[possiblePath].add_supported_op(HydraClassOp(op_name,
+                                                                                    op_method,
```
chrizandr: Methods are lower case in OpenAPI, but in Hydra, the standard is to use upper case method names. Convert "put"/"post"/"get"/"delete" to "PUT"/"POST"/"GET"/"DELETE". This is also messing up the proper `@type` for the supported operations.

vaibhavchellani: yeah my bad , will correct this .

hydrus/parser/openapi-hydra_parser.py
```diff
@@ -0,0 +1,121 @@
...
+            desc = definitions[class_]["description"]
+        except KeyError:
+            desc = class_
+        classDefinition = HydraClass(class_, class_, desc, endpoint=False)
+        properties = definitions[class_]["properties"]
+        classAndClassDefinition[class_] = classDefinition
+        definitionSet.add(class_)
+        api_doc.add_supported_class(classDefinition, collection=False)
+        for prop in properties:
+            new_prop = HydraClassProp("vocab:" + prop, prop, required=False, read=True, write=True)
```
Shekharrajak: maintaining  line length < 80 according to PEP8, will be better.

vaibhavchellani: okay

vaibhavchellani:
@xadahiya @chrizandr @Shekharrajak fixed most of the things pointed out

xadahiya:
Merging #197

PR #208, opened 2018-06-18
Removed Global Variables 

### LINKED ISSUE #203: Single abstract method for OpenAPI parser
### I'm submitting a 
- [ ] bug report.
- [x] feature request.

### Current Behaviour:

In OpenAPI parser, there are global variables such as `classAndClassDefinition` and `definitionSet` in L214 and L215 respectively. These variables are not explicitly declared as global, and are defined under the condition `if __name__ == "__main__"`. In case the methods of the parser are imported from another script, this portion of code would not run, thus causing the declarations/initialisation of the variables to be ignored and causing errors in the code. It is advisable to not use global variables unless it is absolutely necessary. Any user must otherwise create such global variables before he/she actually uses the methods defined in the parser.

### Expected Behaviour:

There should be a much more abstract method that takes an OpenAPI spec(as a json object) and returns a Hydra API Documentation(as a json object). This method must take care of all the necessary declarations and initialisation and must pass variables as parameters, rather than using global declarations. Once the objects have been modified, they can be returned to the parent method using the `return` statement. Please note that you can return multiple objects in using the `return` statement in Python.

No

vaibhavchellani:
Fixes #203 

### Checklist
- [ X] My branch is up-to-date with upstream/develop branch.
- [ X] Everything works and tested for Python 3.5.2 and above.

### Description

created a function so that the parser can be used as a module by other functions , it takes cares of all the issues brought to attention in #203 .
### Test Logs

#### Added

- Feature 1
- Feature 2

#### Changed

- Change 1
- Change 2

#### Fixed

- Bug 1

#### Removed

- Deprecated feature 1

@xadahiya @chrizandr @Mec-iS please review


xadahiya:
@vaibhavchellani  Please resolve the conflicts

vaibhavchellani:
@xadahiya done !

PR #212, opened 2018-06-22
Parsing  the class location iteratively 

vaibhavchellani:
Fixes #211

### Checklist
- [X] My branch is up-to-date with upstream/develop branch.
- [X] Everything works and tested for Python 3.5.2 and above.

### Description

The parser can now go to any level deep to parse class or other objects .Earlier we were only able to go 1 level deep now we are parsing iteratively to fetch data.
### Test Logs
screen shot 2018-06-18 at 1 28 30 pm

PR #214, opened 2018-06-22
removed conditions restricting addition of only classes with paths

### LINKED ISSUE #213: Add all definitions to Api Doc not only definitions which have paths
### I'm submitting a 
- [X] bug report.
- [ ] feature request.

### Current Behaviour:

Currently we are only adding classes which have a path associated with them , but we should have all the defined classes in the doc so that other classes can have semantic refs to those 
### Expected Behaviour:

we will add all the classes with definitions in classAndClassDefinition and definition set in the api doc 

yes

vaibhavchellani:
Fixes #213 

### Checklist
- [X ] My branch is up-to-date with upstream/develop branch.
- [X ] Everything works and tested for Python 3.5.2 and above.

### Description

All the classes/ definitions that have been parsed will be added .
### Test Logs

### DIFF
--- a/hydrus/parser/openapi_parser.py
+++ b/hydrus/parser/openapi_parser.py
@@ -224,17 +224,12 @@ def get_paths(doc: Dict["str", Any], classAndClassDefinition: Dict["str",HydraCl
                     if collection is "true":
                         api_doc.add_supported_class(
                             classAndClassDefinition[class_name], collection=True)
-                        definitionSet.remove(class_name)
                     else:
                         api_doc.add_supported_class(classAndClassDefinition[class_name], collection=False)
-                        definitionSet.remove(class_name)
 
-    for index in range(len(definitionSet)):
-        try:
-            print("start adding"+definitionSet[index])
-            api_doc.add_supported_class(classAndClassDefinition[definitionSet[index]], collection=False)
-        except KeyError:
-            pass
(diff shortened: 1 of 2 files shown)


hydrus/parser/openapi_parser.py
```diff
@@ -219,12 +219,22 @@ def get_paths(doc: Dict["str", Any], classAndClassDefinition: Dict["str",HydraCl
...
+                        api_doc.add_supported_class(classAndClassDefinition[class_name], collection=False)
+                        definitionSet.remove(class_name)
+
+    for index in range(len(definitionSet)):
+        try:
+            print("start adding"+definitionSet[index])
+            api_doc.add_supported_class(classAndClassDefinition[definitionSet[index]], collection=False)
+        except KeyError:
+            pass
     generateEntrypoint(api_doc)
```
xadahiya: You should remove `generateEntrypoint()` from `get_paths()` as it has nothing to do with parsing paths from the doc or am I missing something here?

vaibhavchellani: no you are right , but it was harmless so i left it there , moving it to parse method .

vaibhavchellani:
please merge this after PR #212 so that we can get doc strings and parameter description from that PR.

vaibhavchellani:
@xadahiya fixed

PR #224, opened 2018-06-29
Implementing arrays as inputs parameters 

### LINKED ISSUE #200: Implementing array parameters as input
### I'm submitting a 
- [ ] bug report.
- [ X] feature request.

### Current Behaviour:

currently using hydrus we cannot accept arrays as inputs in requests .
### Expected Behaviour:

we would be able to handle requests like 'endpoint/[variable 1 , variable 2 , variable 3]

yes , definitely

vaibhavchellani:
Fixes #200 

### Checklist

### Description

It allows us to add , update , delete multiple objects in a single request by allowing arrays to be accepted as inputs .

### DIFF
--- a/hydrus/app.py
+++ b/hydrus/app.py
@@ -31,14 +31,18 @@ from flask_restful import Api, Resource
 from flask_cors import CORS
 
 from hydrus.data import crud
-from hydrus.data.exceptions import (ClassNotFound, InstanceExists, PropertyNotFound,
-                                    NotInstanceProperty, NotAbstractProperty,
-                                    InstanceNotFound)
+from hydrus.data.exceptions import (
+    ClassNotFound,
+    InstanceExists,
+    PropertyNotFound,
+    NotInstanceProperty,
+    NotAbstractProperty,
+    InstanceNotFound)
 from hydrus.data.user import check_authorization, add_token, check_token, create_nonce
 from hydrus.utils import get_session, get_doc, get_api_name, get_hydrus_server_url, get_authentication, get_token
 
 from flask.wrappers import Response
-from typing import Dict, List, Any, Union,Optional
+from typing import Dict, List, Any, Union, Optional
 
 
 def validObject(object_: Dict[str, Any]) -> bool:
@@ -51,6 +55,33 @@ def validObject(object_: Dict[str, Any]) -> bool:
     if "@type" in object_:
         return True
     return False
+
+
+def validObjectList(objects_: List[Dict[str, Any]]) -> bool:
+    """
+        Check if the List of Dicts passed are of the valid format or not.
+        (if there's an "@type" key in the dict)
+
+    :param objects_: Object to be checked
+    """
+    for object_ in objects_:
+        if "@type" not in object_:
+            return False
+    return True
+
+
+def type_match(object_: List[Dict[str, Any]], obj_type: str) -> bool:
+    """
+    Checks if the object type matches for every object in list.
+    :param object_: List of objects
+    :param obj_type: The required object type
+    :return: True if all object of list have the right type
+            False otherwise
+    """
+    for obj in object_:
+        if obj["@type"] != obj_type:
+            return False
+    return True
 
 
 def token_response(token: str) -> Response:
@@ -74,12 +105,16 @@ def failed_authentication(incorrect: bool) -> Response:
         message = {401: "Incorrect credentials"}
         realm = 'Basic realm="Incorrect credentials"'
     nonce = create_nonce(get_session())
-    response = set_response_headers(jsonify(message), status_code=401,
-                                    headers=[{'WWW-Authenticate': realm}, {'X-Authentication': nonce}])
+    response = set_response_headers(jsonify(message), status_code=401, headers=[
+                                    {'WWW-Authenticate': realm}, {'X-Authentication': nonce}])
     return response
 
 
-def set_response_headers(resp: Response, ct: str="application/ld+json", headers: List[Dict[str, Any]]=[], status_code: int=200) -> Response:
+def set_response_headers(resp: Response,
+                         ct: str="application/ld+json",
+                         headers: List[Dict[str,
+                                            Any]]=[],
+                         status_code: int=200) -> Response:
     """Set the response headers."""
     resp.status_code = status_code
     for header in headers:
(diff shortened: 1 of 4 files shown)


vaibhavchellani:
Do not review this yet , its work in progress , just created the PR to share the progress and discuss the issues

vaibhavchellani:
the issue with the implementation till now is this , in a `PUT` method the user has to specify the ids of objects , but how will the client know which ids are available , so this has to be done automatically .
Its working well for `POST` requests since if a user does 'DsCollection/1` it updates one object and if the request is `DsCollection/1,2,3` then all objects are updated. 
For `PUT` requests if a client sends `DsCollection/add` with multiple objects in the body , then all the objects are added and ids are assigned , if the user sends `DsCollection/add/1,2,3` then the ids specified are used . But we would have to add  `add` to the URL which i was trying to avoid .

xadahiya:
Okay, couple of things. We have 4 type of operations GET, PUT, POST and DELETE.
1. `GET` - We're designing get so user can fetch multiple objects at once saving some network requests. For this we can have two types of implementation. First is we can simply use `/get` user provides the object IDs as parameters and the server returns the items (no matter what class they belong to) you can use 404 of not found in some object id is invalid in the list. Second is we implement `/SomeCollection/get` here the server only returns the items that belong to that class else invalid or something similar.
Research what we need for openAPI and implement this accordingly.
2. `PUT` - PUT is used to add items to the server. It makes more sense to create a `/add` or `/SomeCollection/add` and user can just provide a list of objects and server adds them and returns the added IDs. If the user wants to custom IDs then he'll need to add items manually and we already have a mechanism in place for that. So, I don't think user needs to provide IDs in this case.
3. `POST` is for updating items. I don't think we'll need to implement this as users can update each object manually. Doesn't make sense to uodate a lot of objects at once as if there are some errors or users send some request with incorrect data then a lot of data will be lost and these changes are non reversible.
4. `DELETE` - This can simply be implemented as get. User provides some ids and we use endpoint like `delete` or `SomeCollection/delete` and the server deletes those objects.

Please check some open API examples and check with of the two syntaxes will be more suitable `/SomeCollection/get` or simple `get`. Same goes for other endpoints as well.

@vaibhavchellani  sorry for being not very clear earlier. Let me know if I missed something.

vaibhavchellani:
@xadahiya thank you for the detailed response , about the post request with multiple objects i think there can be use cases for that functionality ,  for example an admin wanting to change the ids of all user accounts from integer to string . I think we dont need to choose between `someCollection/get` or 'get` , as you said if a user wants to fetch a particular object its possible by `someCollection/id` if the requirement is for all objects simply `GET` on `someCollection` .

xadahiya:
>  ids of all user accounts from integer to string

It's highly unlikely and most of the open API specs won't have such kind of needs. Don't you think?
We should be focusing on the Open API part right now.

> I think we dont need to choose between someCollection/get or get , as you said if a user wants to fetch a particular object its possible by someCollection/id if the requirement is for all objects simply GET on someCollection

:+1:

vaibhavchellani:
@xadahiya please review when free and let me know if any changes have to be made

hydrus/app.py
```diff
@@ -535,8 +761,10 @@ def app_factory(API_NAME: str="api") -> Flask:
                      "/<string:path>", endpoint="item_collection")
     api.add_resource(Item, "/" + API_NAME +
                      "/<string:path>/<int:id_>", endpoint="item")
-    api.add_resource(Items,"/"+ API_NAME +
-                     "/<string:path>/list",endpoint="items")
+    api.add_resource(Items, "/" + API_NAME +
+                     "/<string:path>/add/<int_list>", "/" + API_NAME +
```
xadahiya: I believe it should be `/get` instead of `/add`.

hydrus/app.py
```diff
@@ -467,24 +546,176 @@ def delete(self, path: str) -> Response:
...
+                                jsonify(response), headers=headers_, status_code=201)
+                        except (ClassNotFound, InstanceExists, PropertyNotFound) as e:
+                            status_code, message = e.get_HTTP()
+                            return set_response_headers(
+                                jsonify(message), status_code=status_code)
+
+                return set_response_headers(
+                    jsonify({400: "Data is not valid"}), status_code=400)
+
+            elif path in get_doc().parsed_classes and path + "Collection" not in get_doc().collections:
```
xadahiya: If it's not a collection endpoint then the user shouldn't be able to insert multiple items. This should return this error instead.

hydrus/app.py
```diff
@@ -467,24 +546,176 @@ def delete(self, path: str) -> Response:
...
+                            status_code, message = e.get_HTTP()
+                            return set_response_headers(
+                                jsonify(message), status_code=status_code)
+
+                return set_response_headers(
+                    jsonify({400: "Data is not valid"}), status_code=400)
+
+        abort(endpoint_['status'])
+
+    def post(self, path, int_list):
```
xadahiya: We don't have any endpoint for `POST`. Also, like we discussed we should skip the updating multiple parts for now.

hydrus/data/crud.py
```diff
@@ -210,6 +223,163 @@ def insert(object_: Dict[str, Any], session: scoped_session, id_: Optional[int]
...
+
+    # Try inserting new object
+    try:
+        insert_multiple(objects_=objects_, id_=ids_string, session=session)
+    except (ClassNotFound, InstanceExists, PropertyNotFound) as e:
+        # Put old object back
+        insert_multiple(object_=instances, id_=ids_string, session=session)
+        raise e
+    for id_ in ids_:
+        get(id_=id_, type_=type_, session=session, api_name=api_name, path=path)
```
xadahiya: Isn't this loop redundant? If you're returning `ids_` then why do we need this?

xadahiya:
> I think we dont need to choose between someCollection/get or get , as you said if a user wants to fetch a particular object its possible by someCollection/id if the requirement is for all objects simply GET on someCollection

What if the user wants to fetch certain items from a collection like `/someCollection/get/1,2,3?

hydrus/app.py
vaibhavchellani: but its being used to add objects right ?

hydrus/app.py
vaibhavchellani: yes you are right , i missed that

hydrus/app.py
vaibhavchellani: yes ,removing

hydrus/app.py
xadahiya: I haven't run it yet.

hydrus/data/crud.py
vaibhavchellani: well yeah it is , ill remove that , and should i remove [this[(https://github.com/httP-APIs/hydrus/blob/develop/hydrus/data/crud.py#L283) too ?

xadahiya: Okay, got it! My bad sorry. This is being used to check if objects are inserted successfully. If get fails that'll mean objects were not inserted successfully and exceptions will be raised accordingly.

vaibhavchellani: oh okay

xadahiya:
Take a look at some OpenAPI spec examples and if you think this needs to be implemented then implement it accordingly.

vaibhavchellani:
@xadahiya i made the changes , will open another PR for get , for now it satisfies most of the use cases

xadahiya:
Update the tests then we can merge this.

hydrus/app.py
```diff
@@ -535,8 +694,10 @@ def app_factory(API_NAME: str="api") -> Flask:
                      "/<string:path>", endpoint="item_collection")
     api.add_resource(Item, "/" + API_NAME +
                      "/<string:path>/<int:id_>", endpoint="item")
-    api.add_resource(Items,"/"+ API_NAME +
-                     "/<string:path>/list",endpoint="items")
+    api.add_resource(Items, "/" + API_NAME +
+                     "/<string:path>/add/<int_list>", "/" + API_NAME +
+                     "/<string:path>/add", "/" + API_NAME +
```
xadahiya: What will this path do?

vaibhavchellani: this basically allows us to redirect routes '/add' and '/add/1,2,3' to the same class

xadahiya:
@vaibhavchellani  Have you added tests? We need to merge this

xadahiya:
Merging #224

PR #232, opened 2018-07-07
Implementing new hydrus features in parser 

### LINKED ISSUE #229: Implement New features of hydrus in parser
### I'm submitting a 
- [ ] bug report.
- [X] feature request.

### Current Behaviour:

Right now the parser is only able to handle a very small subset of possible endpoints in OAS , we have added some great functionality to Hydrus which would allow the parser to increase the size of the subset . 
### Expected Behaviour:

We would be able to parse a greater amount of possible endpoints

yes

vaibhavchellani:
Fixes #229 

### Checklist

### Description

New design allows the parser to be updated easily , it allows us to define rules at various places so that we can just update the rules when we update hydrus thus making it very easy to implement updates . The global state allows less fuss and is efficient way to store data. Implemented new hydrus features like custom endpoint .
### Test Logs

#### Added

- Feature 1
- Feature 2


hydrus/parser/openapi_parser.py
```diff
@@ -114,103 +217,196 @@ def get_class_details(class_location: List[str],
...
+                elif type == "object":
+                    print("parameter type object !!")
+                    print("converted to string")
+                    param = "string"
+                else:
+                    param = type_ref_mapping(type)
+
+        else:
+            print("cannot process the parameter")
+            print(parameter)
```
xadahiya: Shouldn't this be inside a try-except block?

vaibhavchellani: im not expecting any exceptions , i just have to check if the param block follows the rules we have defined in allow_paramter

PR #249, opened 2018-07-17
Agent UUID

xadahiya:
Add uuid() implementation to develop


xadahiya:
@vaibhavchellani  Please fix these conflicts so we can use the UUID mechanism in the demo.

vaibhavchellani:
@xadahiya on it

PR #251, opened 2018-07-19
Generalised tests

vaibhavchellani:
Fixes #246 

### Checklist
- [X] My branch is up-to-date with upstream/develop branch.
- [X] Everything works and tested for Python 3.5.2 and above.

### Description

This PR created a generalised test suite for testing any hydra doc for crud operations 
### Test Logs

screen shot 2018-07-19 at 3 14 38 pm

PR #252, opened 2018-07-20
created generalised tests

vaibhavchellani:
In ref to #251

### DIFF
--- a/hydrus/tests/test_crud_general.py
+++ b/hydrus/tests/test_crud_general.py
@@ -1,0 +1,243 @@
+"""Unit tests for CRUD operations in hydrus.data.crud."""
+
+import unittest
+
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker, scoped_session
+import hydrus.data.crud as crud
+from hydrus.data.db_models import Base
+from hydrus.data import doc_parse
+from hydrus.hydraspec import doc_maker
+from hydrus.samples.hydra_doc_sample import doc as api
+import random
+from typing import List
+import string
+
+
+def gen_dummy_object(class_, doc):
+    """Create a dummy object based on the definitions in the API Doc."""
+    object_ = {
+        "@type": class_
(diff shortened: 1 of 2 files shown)


hydrus/tests/test_crud_general.py
```diff
@@ -0,0 +1,243 @@
...
+        id_list.pop()
+        id_list.pop()
+        host = '/'.join(id_list)[0:]
+        self.HYDRUS_SERVER_URL = host+'/'
+        self.API_NAME = api["@id"].split('/')[len(api["@id"].split('/')) - 2] + '/'
+        self.session = session
+        self.doc = doc_maker.create_doc(
+            api, self.HYDRUS_SERVER_URL, self.API_NAME)
+        test_classes = doc_parse.get_classes(api)
+        self.test_class = test_classes[0]["title"]
```
xadahiya: You should use random function to get the class title from list. That'll make tests more generalized.

xadahiya: @vaibhavchellani ^^ We need to merge this.

vaibhavchellani: sorry totally missed this 
updating

vaibhavchellani: @xadahiya the test are failing in some cases where we randomly get collections instead of class , the tests are made for classes only i think , because if its a collection it wont pass the condition in gen_dummy_object

xadahiya: @vaibhavchellani fix this like we discussed yesterday, then we can merge this.

xadahiya: @vaibhavchellani Please fix this, we need to merge.

xadahiya: @vaibhavchellani Please fix this, we need to merge.

vaibhavchellani:
creating new PR

PR #259, opened 2018-08-05
addde conditional statement to check open api doc

### LINKED ISSUE #256: Create a cli command for parser
### I'm submitting a 
- [ ] bug report.
- [X] feature request.

### Current Behaviour:

Right now we dont have a cli command for parser 
### Expected Behaviour:

User will just give the location of open api doc and the command will start the server .

yes

vaibhavchellani:
Fixes #256 

### Checklist

### Description

Creates a CLI command for parser , user has to use `-o` flag and supply location of open api doc , the command parses the open api doc at location , creates a hydra doc and starts a server for the hydra doc .
### Test Logs

screen shot 2018-08-04 at 6 19 20 pm

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -37,7 +37,7 @@ import yaml
               help="Toggle token based user authentication.")
 @click.option("--serverurl", default="http://localhost",
               help="Set server url", type=str)
-@click.option("--openapi","-o" ,default="./hydrus/samples/petstore_openapi.yaml",type=click.File('r'),help="Location to Open API doc")
+@click.option("--openapi","-o" ,default="doc.jsonld",type=click.File('r'),help="Location to Open API doc")
 @click.argument("serve", required=True)
 def startserver(adduser: Tuple, api: str, auth: bool, dburl: str,
                 hydradoc: str, port: int, serverurl: str, token: bool,
@@ -62,15 +62,16 @@ def startserver(adduser: Tuple, api: str, auth: bool, dburl: str,
     # DB_URL = 'sqlite:///database.db'
     DB_URL = dburl
 
-    with open(openapi.name, 'r') as stream:
-        try:
-            openapi_doc = yaml.load(stream)
-        except yaml.YAMLError as exc:
-            print(exc)
-    api_doc = parse(openapi_doc)
-    f = open("./hydrus/samples/hydra_doc_sample.py", "w")
-    f.write(api_doc)
-    f.close()
+    if openapi.name != "doc.jsonld":
+        with open(openapi.name, 'r') as stream:
+            try:
+                openapi_doc = yaml.load(stream)
+            except yaml.YAMLError as exc:
+                print(exc)
+        api_doc = parse(openapi_doc)
+        f = open("./hydrus/samples/hydra_doc_sample.py", "w")
+        f.write(api_doc)
+        f.close()
 
     # Define the server URL, this is what will be displayed on the Doc
     HYDRUS_SERVER_URL = "{}:{}/".format(serverurl, str(port))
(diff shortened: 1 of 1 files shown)


cli.py
```diff
@@ -32,6 +32,9 @@
               help="Toggle token based user authentication.")
 @click.option("--serverurl", default="http://localhost",
               help="Set server url", type=str)
+
+@click.option("--openapi","-o" ,default="doc.jsonld",type=click.File('r'),help="Location to Open API doc")
```
xadahiya: Why would default OpenAPI file be in `doc.jsonld` format?

cli.py
```diff
@@ -32,6 +32,9 @@
...
+
 @click.argument("serve", required=True)
 def startserver(adduser: Tuple, api: str, auth: bool, dburl: str,
                 hydradoc: str, port: int, serverurl: str, token: bool,
```
xadahiya: You haven't passed the openapi parameter here

cli.py
vaibhavchellani: its just a placeholder , i needed default to point to a file , im using it as a flag

cli.py
vaibhavchellani: im sorry i dont understand , could you please elaborate ?

xadahiya:
@vaibhavchellani did you test this. I don't think your code is making much sense.

cli.py
```diff
@@ -56,6 +59,16 @@ def startserver(adduser: Tuple, api: str, auth: bool, dburl: str,
...
     DB_URL = dburl
 
+    if openapi.name != "doc.jsonld":
+        with open(openapi.name, 'r') as stream:
+            try:
+                openapi_doc = yaml.load(stream)
+            except yaml.YAMLError as exc:
+                print(exc)
+        api_doc = parse(openapi_doc)
+        f = open("./hydrus/samples/hydra_doc_sample.py", "w")
```
xadahiya: Why do you need to save api_doc? Just use this parsed doc to start the server.
Ideally, there should be two commands one to convert and save openapi doc to hydradoc and other to directly serve using openapi doc.

cli.py
xadahiya: In the startserver function, all the click parameters need to be passed. Don't you think?

cli.py
vaibhavchellani: the api doc is string , like [this](https://github.com/HTTP-APIs/hydrus/blob/develop/hydrus/samples/hydra_doc_sample.py) , i need the doc object from the api doc thats why i saved it and later used the doc imported .

cli.py
vaibhavchellani: yes its weird i added it , but somehow its not here , sorry my bad pushing

vaibhavchellani:
yes i agree its a bit non conventional and maybe i am doing something wrong (code design wise)  , let me explain a bit
1. The default value in the click command is being used as a flag, if its anything other than doc.jsonld it means user has given the location of open api  and we generate hydra doc and use that doc .
2. I wanted to use the doc generated directly but since it was a string i could not extract the doc object, so i simply save it at a location and pull the doc object via import statement from there, please suggest a better alternative if you know of .

cli.py
xadahiya: Use the parse function from the openapi_parser then

vaibhavchellani: thats what im calling in line 68 right ?

xadahiya: Okay, it's fine then but I you get some time there should be a function to return the doc in proper format not the string then this function can later be used to create JSON dump and string format to store the doc in a file.

vaibhavchellani: yes makes sense ill do that for sure

cli.py
xadahiya: Vaibhav default should be None or something, not the default JSON-LD format. I understand you've very less time left but you should always maintain the coding standards.

vaibhavchellani: i am using click.file hence it has to be a file otherwise its giving error , now i could create another file but it would be hard to explain its existence as just a placeholder .

xadahiya: yes, that'll work.

cli.py
```diff
@@ -56,6 +62,18 @@ def startserver(adduser: Tuple, api: str, auth: bool, dburl: str,
     # DB_URL = 'sqlite:///database.db'
     DB_URL = dburl
 
+    if openapi.name != './petstore_openapi.yaml':
+        print("inside if")
```
xadahiya: @vaibhavchellani Please remove extra print statements.

PR #277, opened 2018-10-02
cli.py cleanup, removed openapi arguement

xadahiya:
The `openapi` argument in `cli.py` was defeating the purpose of having a CLI i.e you can run the command from anywhere but the hydrus main directory. 

**Work on openapi command in progress**


xadahiya:
Moved to #279

PR #295, opened 2018-11-28
Segregate nested objects

### LINKED ISSUE #248: Segregate objects types in responses
Update hydrus to fix https://github.com/HTTP-APIs/hydra-flock-demo/issues/34

vddesai1871:
Fixes #248 

### Checklist

### Description

Segregates objects returned in a response

### Test Logs

```
(h) vishal@vishal:~/oss/hydrus$ pytest hydrus/tests/
============================================================ test session starts ============================================================
platform linux -- Python 3.5.2, pytest-3.4.2, py-1.5.2, pluggy-0.6.0
rootdir: /home/vishal/oss/hydrus, inifile:
collected 51 items                                                                                                                          

hydrus/tests/test_app.py ................                                                                                             [ 31%]
hydrus/tests/test_auth.py ..........                                                                                                  [ 50%]
hydrus/tests/test_crud.py ............                                                                                                [ 74%]
hydrus/tests/test_doc_writer.py ....                                                                                                  [ 82%]
hydrus/tests/test_parser.py .........                                                                                                 [100%]

========================================================= 51 passed in 3.50 seconds =========================================================
(h) vishal@vishal:~/oss/hydrus$ 
```

### Changes log
Added a new resource names 'NestedItem' in app.py which handles the requests for nested objects
#### Added

- Feature 1
Segregates objects returned in a response.
Only segregates the objects nested at the first level, if the nested object has other nested objects in it then they will be returned in the same response.


vddesai1998:
Response for Drone object
drone_ex
Response for State object
state_ex

vddesai1998:
Should we add `POST` method in `NestedItem` so clients can modify the particular nested object directly?
Regarding to changes in flock-demo documentation, some primarily nested classes like `State` do not have any operations defined, I should define `GET` as supported operation for those classes and check in app.py if the nested class supports the `GET` operation before returning it. What other chnages need to be made in ApiDoc?

chrizandr:
https://github.com/HTTP-APIs/hydra-flock-demo/issues/34#issuecomment-445454889

chrizandr:
As far as I can understand, if there are mutliple nestings the old method of retrieving objects is followed right?

Please read the comment I posted, I think we can handle multiple nestings and single level nestings using only the modifications in `crud`, we might not need to add new endpoints for it.

Mec-iS:
This improvement may save different reads as in the place of the full object should appear only its link to the proper URI. Any Hydra Resource has by definition its own endpoint, so `State` does.

vddesai1998:
dc

sc

vddesai1998:
```
(h) vishal@vishal:~/oss/hydrus$ pytest hydrus/tests
============================================================ test session starts ============================================================
platform linux -- Python 3.5.2, pytest-3.4.2, py-1.5.2, pluggy-0.6.0
rootdir: /home/vishal/oss/hydrus, inifile:
collected 51 items                                                                                                                          

hydrus/tests/test_app.py ................                                                                                             [ 31%]
hydrus/tests/test_auth.py ..........                                                                                                  [ 50%]
hydrus/tests/test_crud.py ............                                                                                                [ 74%]
hydrus/tests/test_doc_writer.py ....                                                                                                  [ 82%]
hydrus/tests/test_parser.py .........                                                                                                 [100%]

========================================================= 51 passed in 3.67 seconds =========================================================
(h) vishal@vishal:~/oss/hydrus$ 
```

vddesai1998:
Other classes besides `State` in flock-demo-api-doc, which can be returned as nested object are `Datastream` and `Command`, both already have collections defined over them and support GET requests. So I have to just add collection for `State` and define GET operation for `State` objects in flock-demo-vocab, right?

Mec-iS:
Every object that is a `Resource` and it's present inside another `Resource` should be referencing via its URI. So yes for `Datastream` and `Command` in flock-demo and in general for any other object that is instance of `HydraResource`. 
This is the defining part of an hypermedia API, nested objects have to be accessed via URI dereferencing.

vddesai1998:
>This is the defining part of an hypermedia API, nested objects have to be accessed via URI dereferencing.

Yes I know that, and maybe that's why we prefer HTTP-APIs as org name instead of REST-APIs cause REST constraint of hypermedia as the engine of application state is not fulfilled by many self-proclaimed RESTful APIs. 
As you can see from the screenshot now server just returns a URI for the nested object, I have also checked it for other examples as well. So the changes needed to be made in the server are done and I have to update the flock-vocab in `hydra-flock-vocab` repo.
My question was about the changes regarding to flock-vocab. The repo for drone vocab [hydra-flock-vocab](https://github.com/HTTP-APIs/hydra-flock-vocab/blob/master/hydra_flock.jsonld) has different vocab than one available in the `hydrus/examples/drones` folder in this repo. Different in the sense that it doesn't define classes like `Area`, `LogEntry`, `Message` and their collections. Why is this the case?

Mec-iS:
@vddesai1871 I dont see tests for this feature in th ePR.

vddesai1871:
Added tests and ran tests with both sample doc and drone flock doc from `examples/drones/doc.py`

vddesai1871:
Initially I was only handling collection classes as nested objects. After following your comment 
>Every object that is a Resource and it's present inside another Resource should be referencing via its URI.

Now code also handles non collection classes which may be present at nested classes in api_doc.
Extended sample doc to have both collection class and non collection class as nested classes (property).
Wrote tests and ran with sample doc and flock-drone doc.

PR #299, opened 2018-12-04
Adapt and modify tests

### LINKED ISSUE #296: Adapt and modify tests in test_app.py
### I'm submitting a 
- [x] bug report.
- [ ] feature request.

### Current Behaviour:

Currently, tests in test_app.py are not testing(asserting) the behaviour of the app correctly. For most of the tests, it is not even executing the code which is supposed to test the functionality.
1) In test_Collection_GET and few other tests, we are treating the endpoint as collection name instead of extracting collection name from the endpoint.
2) In test_put_at_id: we are using integers instead of using `uuid`s to test the app
3) In test_put_at_ids: we are using wrong `url` to make the request
4) regular expression to fetch ID is outdated, we should use regex for uuid.
5) Some methods like `DELETE` for collection class is not defined in the sample document we are using, so it should be added.
6) Same goes for class with endpoints, they should be added to the documentation.
### Expected Behaviour:

Every test should be executed.
Tests should use a general documentation we are using right now but with all the functionality supported by `hydrus`, so it can be tested properly. we can update the ApiDoc whenever we add a new functionality to `hydrus`.
### Steps to reproduce:

Try to make a test fail by a false assertion in test_app.py starting from test_Collection_GET, it will still pass because the assertion statements are not being executed for most of the tests.

YES

vddesai1871:
Fixes #296 

### Checklist

### Description

Adapts the tests in test_app.py to recent changes.
### Test Logs

```
(h) vishal@vishal:~/oss/hydrus$ pytest hydrus/tests/
============================================================ test session starts ============================================================
platform linux -- Python 3.5.2, pytest-3.4.2, py-1.5.2, pluggy-0.6.0
rootdir: /home/vishal/oss/hydrus, inifile:
collected 51 items                                                                                                                          

hydrus/tests/test_app.py ................                                                                                             [ 31%]
hydrus/tests/test_auth.py ..........                                                                                                  [ 50%]
hydrus/tests/test_crud.py ............                                                                                                [ 74%]
hydrus/tests/test_doc_writer.py ....                                                                                                  [ 82%]
hydrus/tests/test_parser.py .........                                                                                                 [100%]

========================================================= 51 passed in 3.88 seconds =========================================================
(h) vishal@vishal:~/oss/hydrus$ git add hydrus/samples/doc_writer_sample.py
```
### Changes log
Debugs various errors mentioned in #296 
#### Added

- Feature 1
- Feature 2

#### Changed

- Change 1
- Change 2

#### Fixed

- Bug 1

#### Removed

- Deprecated feature 1

### DIFF
--- a/hydrus/samples/doc_writer_sample.py
+++ b/hydrus/samples/doc_writer_sample.py
@@ -8,6 +8,7 @@ API_NAME = "api"                # Name of the API, will serve as EntryPoint
 BASE_URL = "https://hydrus.com/"    # The base url at which the API is hosted
 # NOTE: The API will be accessible at BASE_URL + ENTRY_POINT (http://hydrus.com/api/)
 
+"""Create ApiDoc Object"""
 api_doc = HydraDoc(API_NAME,
                    "Title for the API Documentation",
                    "Description for the API Documentation",
@@ -19,8 +20,20 @@ api_doc = HydraDoc(API_NAME,
 class_uri = "dummyClass"      # URI of class for the HydraClass
 class_title = "dummyClass"                      # Title of the Class
 class_description = "A dummyClass for demo"     # Description of the class
-
 class_ = HydraClass(class_uri, class_title, class_description, endpoint=False)
+
+"""Class with single instance"""
+class_2_uri = "singleClass"
+class_2_title = "singleClass"
+class_2_description = "A non collection class"
+class_2 = HydraClass(class_2_uri, class_2_title, class_2_description, endpoint=True)
+
+"""Class not having any methods except put and get"""
+class_3_uri = "extraClass"
+class_3_title = "extraClass"
+class_3_description = "Class without any explicit methods"
+class_3 = HydraClass(class_3_uri, class_3_title, class_3_description, endpoint=False)
+
 # NOTE: Setting endpoint=True creates an endpoint for the class itself, this is usually for classes that have single instances.
 #       These classes should not ideally have a Collection, although Hydrus will allow creation of Collections for them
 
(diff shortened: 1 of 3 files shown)


chrizandr:
It would be good if you could also try and run this test for another API documentation, other than the dummy one created in the sample. Maybe try for the flock-demo documentation and see if all tests pass.

vddesai1998:
>Maybe try for the flock-demo documentation and see if all tests pass.
2 tests are failing because I am assuming if an endpoint class has POST or DELETE or GET method it must have a PUT method, so I am using put(directly without checking if class supports PUT operation) to add a new object before testing for getting, posting or deleting it.
(Area class in flock-demo is an endpoint class and doesn't have PUT method defined but has POST defined) 
This piece of code is being used for testing endpoint class POST requests

```
                    if "POST" in class_methods:
                        dummy_object = gen_dummy_object(class_.title, self.doc)
                        put_response = self.client.put(
                            endpoints[class_name], data=json.dumps(dummy_object)
                        )
                        assert put_response.status_code == 201
                        post_response = self.client.post(
                            endpoints[class_name], data=json.dumps(dummy_object))
                        assert post_response.status_code == 200
```

Mec-iS:
it is not mandatory for any endpoint to have other methods beside the ones it declares in the doc. So the test should be fixed to test only the methods allowed by the endpoint, the others should return `Method Not Allowed`.

vddesai1998:
>it is not mandatory for any endpoint to have other methods beside the ones it declares in the doc. So the test should be fixed to test only the methods allowed by the endpoint,

But before checking for POST or DELETE or GET method, don't we have to add an item through PUT so we can update or delete or get it.
If in tests we try to POST an item directly it will give `404`

Mec-iS:
in the context of a unit test, every test case should be absolutely independent from other tests. So in this case, and any similar cases, we should insert the item using the ORM and then test only the `GET`:
```
def test_get_method(self):
     # test setup
     Item.create(...)
     
     # assertion
     assert requests.get(...).status_code == 200
```

Even better: the item creation should be in the class' `setUp` method (check `unittest.TestCase.setUp` in standard library) and reused by all the tests in the class.

hydrus/tests/test_app.py
```diff
@@ -159,75 +162,83 @@ def test_Collections_PUT(self):
         index = self.client.get("/" + self.API_NAME)
         assert index.status_code == 200
         endpoints = json.loads(index.data.decode('utf-8'))
-        for collection_name in endpoints:
+        for endpoint in endpoints:
+            collection_name = "/".join(endpoints[endpoint].split(
```
Mec-iS: it is better to format URLs:
* or using `str.format()`: `'{}/{}/{}'.format(..., ..., ...)`
* or better string formatting: `f'{domain}/{path}/{id}'`

This is a wider problem in the codebase, and refactor should apply. I will open a issue for "General Refactoring".

vddesai1998:
> It would be good if you could also try and run this test for another API documentation, other than the dummy one created in the sample. Maybe try for the flock-demo documentation and see if all tests pass.

Yes now tests are passing for flock-demo doc. Thanks!

hydrus/samples/doc_writer_sample.py
```diff
@@ -58,17 +71,46 @@
                    op_returns,
                    op_status)
 
-
-"""Add the operation to the Class"""
+"""Same way add DELETE, PUT and GET operations"""
```
xadahiya: @vddesai1871  Looks good! Could you please replace the `"""` with simple comments.

xadahiya:
Also, can you fix this deprecation warning? 
annotation 2018-12-19 153045

vddesai1871:
>Also, can you fix this deprecation warning?

I don't think I can do anything about this. It is related to pyyaml's compatibility with pythin 3.7 [they are woking on it](https://github.com/yaml/pyyaml/pull/220)

PR #309, opened 2018-12-22
Add expiry field to Token

### LINKED ISSUE #178: Expired token should be removed from the database

invinciblycool:
Fixes #178

### Checklist

### Description

 - The PR adds a `expiry` field to Token with an expiration time of 2700 seconds and makes the necessary changes in the dependent functions.
 - A possible addition could be adding an option in the `cli` to set the expiration time manually.

Although, the best approach here should be using JWT which encodes user info, coupled with redis, since using a RDBMS would unintentionally add write overhead.

### Test Logs

```
========================== test session starts ===========================
platform linux -- Python 3.6.7, pytest-4.0.2, py-1.7.0, pluggy-0.8.0
rootdir: /home/invinciblycool/Documents/Github-projects/hydrus, inifile:
collected 51 items                                                       

hydrus/tests/test_app.py ................                          [ 31%]
hydrus/tests/test_auth.py ..........                               [ 50%]
hydrus/tests/test_crud.py ............                             [ 74%]
hydrus/tests/test_doc_writer.py ....                               [ 82%]
hydrus/tests/test_parser.py .........                              [100%]

======================= 51 passed in 1.42 seconds ========================
```

### DIFF
--- a/hydrus/data/db_models.py
+++ b/hydrus/data/db_models.py
@@ -3,13 +3,17 @@
 from sqlalchemy import create_engine, ForeignKey
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy import Column, Integer, String, DateTime
+from sqlalchemy.sql import func
 from typing import Any
+import datetime
 import uuid
 # from hydrus.settings import DB_URL
 
 engine = create_engine('sqlite:///database.db')
 
 Base = declarative_base()  # type: Any
+
+EXPIRY_TIME = 2700  # unit: seconds
 
 
 class RDFClass(Base):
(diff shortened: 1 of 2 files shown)


Mec-iS:
looks legit.

Please @chrizandr or @xadahiya take a look and merge eventually.

PR #310, opened 2018-12-27
Merge latest develop to master

Mec-iS:


xadahiya:
@Mec-iS I think we should merge it now.

Issue #317, opened 2018-12-31
function to generate dummy objects fails for classes with explicitly defined paths.


vddesai1871:
### I'm submitting a 
- [x] bug report.
- [ ] feature request.

### Current Behaviour:

This function is used to create dummy classes for test purposes.
```
def gen_dummy_object(class_, doc):
    """Create a dummy object based on the definitions in the API Doc."""
    object_ = {
        "@type": class_
    }
    if class_ in doc.parsed_classes:
        for prop in doc.parsed_classes[class_]["class"].supportedProperty:
            if "vocab:" in prop.prop:
                prop_class = prop.prop.replace("vocab:", "")
                object_[prop.title] = gen_dummy_object(prop_class, doc)
            else:
                object_[prop.title] = ''.join(random.choice(
                    string.ascii_uppercase + string.digits) for _ in range(6))
        return object_
```
The first parameter of the function `class_` is ambiguous. Its not clear whether it should take class.title as its value or class.path as its value. In either case, it will fail for classes with explicitly given path names.
Suppose the class for which we want to create dummy objects has explicitly given path.
1) And suppose the `class_` parameter takes class.title as value and use that to create dummy object, 
then ` if class_ in doc.parsed_classes:` will return false. And hence the dummy object won't be created successfully.
2) now suppose the `class_` parameter takes class.path as value, 
then 
```
 object_ = {
        "@type": class_
    }
```
will create an invalid object as the `@type` property must have class.tilte as its value not class.path.
### Expected Behaviour:

From the parameter name, it should be clear that it takes class_path (or class_title, either will work) as value and then inside the function we can use it to retrieve class.title and set `@type` correctly, anything else won't need to be changed.
(Apologies for the long explanation, it was really an insidious bug ;))
### Steps to reproduce:

Create an apiDoc having class with explicitly defined path, and try to run tests or run `gen_dummy_obj()` function
### Do you want to work on this issue?

Yes

Mec-iS:
I think `class_` is a HydraClass object not a `str`. I agree that parameters shall be documented and their types explicitly documented a well.

vddesai1871:
>I think class_ is a HydraClass object not a str.

From function body and function calls it is clear we are expecting `class_` to be a string. But it was not clear whether it is `class.title` or `class.path`. We were passing `class.title` as value in function calls which is bound to fail when the class has explicitly defined path. I was suggesting to use class_title or class_path as the parameter and change the code accordingly. 
[...]

PR #326, opened 2019-01-05
Implement read-only, write-only and required properties

### LINKED ISSUE #226: Implementing readOnly, writeOnly and Required Properties
### I'm submitting a 
- [ ] bug report.
- [X] feature request.

### Current Behaviour:

Right now although the doc contains properties like readOnly ,writeOnly , Required the functionality is not implemented in Hydrus. We need to implement this functionality so that we can have checksums instead of pk's or UUID's , advantage of that would be that client can reconstruct the checksum to check what is to be fetched .

~YES~

EDITED BY Mec-iS

vddesai1871:
Fixes #226 

### Checklist

### Description

Implemented read-only, write-only and required properties.
### Change logs

Added code to handle read-only, write-only and required properties.
Added tests for write-only and required-props. 

Adding tests for read-only props requires major changes in test file, which involves #317 and other changes to generations of dummy objects respecting read-only constraint and create appropriate objects of PUT (should have read-only props) and POST(should not have read-only props), and adapting to those changes in almost every individual test. So I will add that in another PR after discussing. (We have #317 open so we won't miss out adding this test).


chrizandr:
Will check this out tomorrow in detail.

vddesai1871:
>A better approach would be to store read-only, write-only and required values along with the property definitions which are created when the initial db schema is generated and all the classes and properties are added to the database.

I had to pass api_doc to crud.py for implementing write-only props. Read-only and required works without any modification in crud.py. So I think we will have to just store write-only pros in db.

But we are also using doc for segragation of objects in responses. 
(ie. 
```
            if doc.collections[collection]["collection"].class_.path == inst_class_name:
                nested_class_path = doc.collections[collection]["collection"].path
```
)
How can we achive this without passing api_doc?

vddesai1871:
>I don't remember how things are implemented is read-only, write-only stored in the database. How can these work but not the required. Can you explain a bit

Read-only and required work without crud.py but write-only don't.
I am handling this constraint as they occur, so If a client tries to create a new object or update an old object ( he must set the value of all required properties in request data(object)), so I am checking that in resources.py (in `put` and `post` methods).
Same goes for read-only properties handled in resources.py.
For write-only properties, the server is supposed to do not add write-only properties and their values in response, and response object is being created in crud.py so I've written code in crud.py to skip fetching and adding write-only properties in the response object.(We can also do this in resources.py by deleting write-only properties from the response object created by crud.py before sending it to the client, but I thought that won't be efficient)

xadahiya:
@Mec-iS  will review in detail and merge over the weekend. This is a crucial part.

xadahiya:
@vddesai1871 Explain me this, why aren't we checking for all 3 `read-only`, `write-only` and `required` types in `crud.py` itself? Whenever tests fail we can just raise an Error and won't need to change anything else. Also, this will make the `param` of all functions in crud equivalent (right now some have doc as parameter and some don't). 
I can't see the benefit of checking `write-only` in crud.py and the rest in resources.py. Things should be consistent don't you think?

vddesai1871:
>I can't see the benefit of checking write-only in crud.py and the rest in resources.py. Things should be consistent don't you think?

@xadahiya I believe handling write-only and other two as the same would be an over-generalization, by a superficial look they do seem closely related but by their essence or by their purpose, they are very different things. Write-only has nothing to do with raising Error or to do anything with request body, it just provides a requirement regarding how the response body will be created( suggesting which properties to exclude from response body). While read-only and required provides information about what is considered as a valid request body and what's not.  And their might be a violation of this desired request body schema and we must handle that by raising some sort of error or exception. **In brief write-only provides constraint on response body and other two provide constriants on request body and that's why they are being handled differently.**

If we were creating a normal 'primitive' API we wouldn't have other choice but to add these fields in db and thus check for them in CRUD, going a step forward we could have used some sort of request serializer like marshmallow to define request-response schema and handle these constriaints at view level. But with hydra we already have api doc (the schema) and we just have to write validation code.

PROPOSAL: Leave crud as it is and handle everything at `view` level in resources.py.
1) Read-only and required will remain same,
2) Create a function to finalize the response data, which will remove write-only data from object returned by crud that's how it (dump-only misnomer for write-only) is being handled in marshmallow and other serializers, the same function would also handle path of nested objects. ( crud will set {`property_name`: id}, the function with help of api_doc will convert it to {`property_name`: actual_path[/id]})
We won't need to pass api_doc to crud, 
Change in api_doc won't require changes in db(If a property's write-only or anyother attribute is changed hydrus will work fine without having to change anything in db).

chrizandr:
@vddesai1871 I agree with your argument, but @xadahiya is right. Both request and responses are handled in crud at the lower levels. The hierarchy is `View -> Crud -> DB`, for both request and response objects. Like you said, we can implement the constraints in `View`, but it can also be implemented in `Crud`. I think we need to choose one only, not have a mix of both. It would be better to have it either in `Crud` or in `View`, but not separate for request and responses.

vddesai1871:
> I think we need to choose one only, not have a mix of both. It would be better to have it either in Crud or in View, but not separate for request and responses.

@chrizandr I agree, that's why I proposed handling write-only also in `View` in my latest proposal.

vddesai1871:
>I have one more idea that I had in mind for a while now, we could modify the API Doc object to do all these checks. Right now, there are a lot of nested ifs in the Views because of a large number of API Doc checks that are happening on the object. What if we have just a simple API_Doc.validate() method that checks all constraints and conditions that are required for an object to conform to the doc. This would remove a lot of the complicated code from views. 

I too had something similar in mind as part of `improvement and refactoring in hydrus` GSoC project :)

xadahiya:
@vddesai1871 the latest proposal sounds good. I think you should proceed with this then we can merge it. 
Regarding the rest of the improvements I think we can do those during GSoC.

Mec-iS:
@xadahiya @chrizandr 
Is this ok to merge?

hydrus/helpers.py
```diff
@@ -103,3 +103,42 @@ def checkClassOp(class_type: str, method: str) -> bool:
         if supportedOp.method == method:
             return True
     return False
+
+
+def check_required_props(class_type: str, obj: Dict[str, Any]) -> bool:
+    for prop in get_doc().parsed_classes[class_type]["class"].supportedProperty:
```
Mec-iS: docstring for the function is missing

hydrus/helpers.py
```diff
@@ -103,3 +103,42 @@ def checkClassOp(class_type: str, method: str) -> bool:
...
+        if prop.required:
+            if prop.title not in obj:
+                return False
+    return True
+
+
+def check_read_only_props(class_type: str, obj: Dict[str, Any]) -> bool:
+    for prop in get_doc().parsed_classes[class_type]["class"].supportedProperty:
```
Mec-iS: docstring for the function is missing

hydrus/helpers.py
```diff
@@ -103,3 +103,42 @@ def checkClassOp(class_type: str, method: str) -> bool:
...
+        if prop.read:
+            if prop.title in obj:
+                return False
+    return True
+
+
+def get_nested_class_path(class_type: str) -> Tuple[str, bool]:
+    for collection in get_doc().collections:
```
Mec-iS: docstrings missing

hydrus/helpers.py
```diff
@@ -103,3 +103,42 @@ def checkClassOp(class_type: str, method: str) -> bool:
...
+        if get_doc().collections[collection]["collection"].class_.title == class_type:
+            return get_doc().collections[collection]["collection"].path, True
+
+    return get_doc().parsed_classes[class_type]["class"].path, False
+
+
+def finalize_response(class_type: str, obj: Dict[str, Any]) -> Dict[str, Any]:
+    for prop in get_doc().parsed_classes[class_type]["class"].supportedProperty:
```
Mec-iS: docstring missing

xadahiya:
We should merge this. @vddesai1871 Can you please resolve the conflicts?

PR #341, opened 2019-01-10
Refactor deeply nested if blocks

### LINKED ISSUE #327: Refactor deeply nested if blocks
### I'm submitting a 
- [ ] bug report.
- [x] feature request.

### Current Behaviour:

Currently the code in `resources.py` has deeply nested if blocks which decrease the readability of the code, also it causes the code to violate PEP max-line-length constraint, trying to avoid that we use line breaks which makes the code even less readable.

### Expected Behaviour:

Instead of using nested Ifs we should check every condition separately and return an appropriate response for that.
Ex.
```
if condition_1:
     if condition_2:
              if condition_3:
                    do_something
              else:
                    response for violation of condition_3
     else
           response for violation of conditon_2
else:
       response for violation of condition_1
```
should become
```
if(!condition_1):
     handle_it
elif(!conditio_2):
     handle_it
elif(!condtion_3):
     handle_it
else:
      do_something
```
So to summarize instead of checking for conditions to be true we should check for false and handle it immediately.

No

de-sh:
Fixes #327 

### Checklist

### Description

Changes source code to un-roll deeply nested if blocks into more readable if-elif ladders. This is on the basis of my understanding of the issue and needs further refinement through discussion.

### Change logs
Refactored deeply nested if blocks in source code.

#### Changed
hydrus/resources.py, refactor deeply nested if blocks.


de-sh:
@vddesai1871, does this commit go in the direction of solving #327 ?

de-sh:
Noted :)

Regards, devduttShenoi<http://devdutt.me>

________________________________

de-sh:
Ok, I'll get to refactoring the rest then!

Mec-iS:
What is the status here?

de-sh:
Almost done

xadahiya:
@Mec-iS devdutt hasn't implemented the changes on all the files. Right now this PR contains only one file that's changed. He's working on implementing the changes across rest of the files.

de-sh:
Hey guys, found a huge bug I had introduced in a former PR of mine that was merged. It's in main right now. Really sorry that happened. I will take care not to let that happen again!

Mec-iS:
@de-sh please take more time to think about problems during your development workflow. you have been very rushy on multiple issues and this does not help sanity checks. Plus anytime something changes there should be a test in place for that part of the code; if it doesn't a test has to be added.

Mec-iS:
@de-sh if there is a problem in `master` you should open a dedicated PR to fix it directly on `master` by prefix the PR with "Fix" so to know that we are fixing directly in the main branch.

de-sh:
@Mec-iS, #346 is to pull the fixes to master.

hydrus/resources.py
```diff
@@ -238,42 +227,39 @@ def get(self, path: str) -> Response:
...
+            except ClassNotFound as e:
+                status_code, message = e.get_HTTP()
+                return set_response_headers(jsonify(message), status_code=status_code)
 
-                except (ClassNotFound, InstanceNotFound) as e:
-                    status_code, message = e.get_HTTP()
-                    return set_response_headers(
-                        jsonify(message), status_code=status_code)
+        # If endpoint and GET method is supported in the API and class is supported
+        elif path in get_doc().parsed_classes and "{}Collection".format(
```
chrizandr: Does this pass PEP8?

hydrus/resources.py
```diff
@@ -298,53 +284,49 @@ def put(self, path: str) -> Response:
...
-                                jsonify(message), status_code=status_code)
+                if validObject(object_) and object_["@type"] == obj_type:
+                    # If Item in request's JSON is a valid object ie. @type is a key in object_
+                    # and the right Item type is being added to the collection
+                    try:
+                        # Insert object and return location in Header
+                        object_id = crud.insert(
+                            object_=object_, session=get_session())
+                        headers_ = [
+                            {"Location": "{}{}/{}/{}".format(
```
chrizandr: Again, why the extra placeholder? This will give an error.

hydrus/data/crud.py
```diff
@@ -211,20 +210,20 @@ def insert(object_: Dict[str, Any], session: scoped_session,
...
+                property_.type_ = "ABSTRACT"
+                session.add(property_)
+                class_ = session.query(RDFClass).filter(
+                    RDFClass.name == object_[prop_name]).one()
+                triple = GraphIAC(
+                    subject=instance.id,
+                    predicate=property_.id,
+                    object_=class_.id)
+                session.add(triple)
             elif session.query(exists().where(RDFClass.name == str(object_[prop_name]))).scalar():
```
chrizandr: Why check twice? DB calls are expensive, try and minimize them.

hydrus/resources.py
chrizandr: Also checking the same condition twice, maybe store in a variable and use? That is why there was nested block, you reduced nesting but there are more calls to the session object. Nesting is okay in favour of speed, more calls to session makes it slower.

de-sh: `Does this pass PEP8?` I know it seems weird, but it does comply :sweat_smile:

hydrus/data/crud.py
de-sh: Ok, that can only be done in a separate PR, could you make an issue and point out what exactly has to be done?

chrizandr: When we had nested `ifs`, the DB condition was only checked once and then the inner conditions are checked. Now because you remove the nesting, we need to check the DB condition for both the inner conditions, this requires 2 DB calls. You can revert the changes or use an intermediate variable that stores the value for the condition. No need for a separate PR though.

de-sh: Ok, that's my fault, I'll get to it as soon as possible 👍

cli.py
```diff
@@ -9,7 +9,7 @@
 from hydrus.data.db_models import Base
 from hydrus.data.user import add_user
 from gevent.pywsgi import WSGIServer
-from hydrus.parser.openapi_parser import parse
+from hydrus_openapi_parser.openapi_parser import parse
```
Mec-iS: This import doesn't look right! @de-sh @sandeepsajan0

PR #346, opened 2019-01-28
FIX: placeholders found missing, placed properly

### LINKED ISSUE #303: Generic Refactoring
### Current Behaviour:
Some code is not readable and some optimization should be applied. For example:
* URL formatting should be using: `str.format()`: `'{}/{}/{}'.format(..., ..., ...)` or better `f'{domain}/{path}/{id}'`
* issue #301
* please @chrizandr @xadahiya add to the list
* Adding doc strings wherever missing

### Expected Behaviour:
Code should be refactored

de-sh:
Some placeholders within str.format() expression were found missing and have been properly placed.

Fixes #303 and is directly on master to fix a bug created in a merged commit.

### Checklist

### Description
This PR is to solve an issue that was brought in due to the lack of placeholders within the str.format() function calls in the resources.py file.

### Change logs
Fix missing {} placeholders within str.format() expressions.

### DIFF
--- a/hydrus/resources.py
+++ b/hydrus/resources.py
@@ -105,13 +105,11 @@ class Item(Resource):
                     api_name=get_api_name(),
                     session=get_session())
 
-                return set_response_headers(
-                    jsonify(hydrafy(response, path=path)))
+                return set_response_headers(jsonify(hydrafy(response, path=path)))
 
             except (ClassNotFound, InstanceNotFound) as e:
                 status_code, message = e.get_HTTP()
-                return set_response_headers(
-                    jsonify(message), status_code=status_code)
+                return set_response_headers(jsonify(message), status_code=status_code)
         abort(405)
 
     def post(self, id_: str, path: str) -> Response:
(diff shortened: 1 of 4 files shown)


hydrus/resources.py
```diff
@@ -143,21 +141,17 @@ def post(self, id_: str, path: str) -> Response:
                             type_=object_["@type"],
                             session=get_session(),
                             api_name=get_api_name())
-                        headers_ = [
-                            {"Location": "{}/{}/{}".format(
+                        headers_ = [{"Location": "{}{}/{}/{}".format(
```
chrizandr: This is wrong. You can't have four placeholders and give only three strings in format. It will raise an IndexError exception.
Why is this being done anyway?

Mec-iS: string formatting is the proper way to do this if we wanto to support Python3.5+.

@de-sh i am starting to dubt that you run tests properly, as these things are quite evident

de-sh: @Mec-iS,

chrizandr: Still don't see what the extra placeholder is for. I am surprised that Python does not give an error for this. Are you running tests on the correct branch, i.e. the one you made the changes to or on the original branch?

vddesai1871: @chrizandr  It looks fine to me
```
headers_ = [{"Location": "{}{}/{}/{}".format(
get_hydrus_server_url(), get_api_name(), path, object_id)}]
```
There are 4 placeholders and 4 strings

xadahiya: Looks like earlier we had 3 placeholders and 4 values (` get_hydrus_server_url(), get_api_name(), path, object_id`). Doesn't seem right. Anyways, the PR looks good to me.

PR #347, opened 2019-01-30
Remove nested if statements

### LINKED ISSUE #327: Refactor deeply nested if blocks

aribalam:
Fixes #327 

### Checklist

### Description
Removes redundant nested if statements from `resources.py`.

### DIFF
--- a/hydrus/resources.py
+++ b/hydrus/resources.py
@@ -132,29 +132,28 @@ class Item(Resource):
             object_ = json.loads(request.data.decode('utf-8'))
             obj_type = getType(class_type, "POST")
             # Load new object and type
-            if validObject(object_):
-                if object_["@type"] == obj_type:
-                    try:
-                        # Update the right ID if the object is valid and matches
-                        # type of Item
-                        object_id = crud.update(
-                            object_=object_,
-                            id_=id_,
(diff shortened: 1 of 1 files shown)


chrizandr:
Seems outdated for now, also there has been no real change in the logic, just nested statements converted to one if statement. 

Closing for now.

PR #350, opened 2019-02-02
Refactoring validation keys in doc_maker

### LINKED ISSUE #348: refactoring in doc_maker
I have a few suggestions - 
1. Currently the `doc_maker` has a number of `doc_keys` defined inside functions which creates classes, properties, operations and statuses to check for those keys. Since those are all hard-coded and are subject to changes, what if we put all those keys in one place as global declarations?
We could then map those keys to their respective `body_type` and use it accordingly.
2. The two functions `class_in_endpoint` and `collection_in_endpoint` has class dictionary as its param which is used only for its title inside the function. We could only pass its title as a param instead of the whole dictionary

aribalam:
Fixes #348

### Checklist

### Description
Currently the hard-coded dictionaries for key validation are present inside function definitions. Being a constant, it should be declared globally according to PEP-8 standards. The PR makes them all global
and creates a function to map them with `body_type`. 

### Change logs
1. Declare the keys as global
2. Create `get_keys` to map those keys with `body_type`
3. Changed and renamed `input_key_check` function to perform all key validations and raise error accordingly

The keys could also be moved to a new file that can serve as a template provider and can be accessed from there.

### DIFF
--- a/hydrus/hydraspec/doc_maker.py
+++ b/hydrus/hydraspec/doc_maker.py
@@ -30,7 +30,7 @@ prop_keys = {
     "required": True
 }
 
-operation_keys = {
+op_keys = {
     "title": False,
     "method": False,
     "expects": True,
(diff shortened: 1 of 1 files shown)


chrizandr:
Those dictionaries were actually templates, the dictionary is created as per the given body. Plus it was not specified in the functions themselves that the variables are global. 

I don't really see why this is better than the previous version of code, right now you have a global and you are creating copies in the function by copying keys and values to a new dictionary, which might be slower, previously we defined them in the function so each function call creates a separate copy. Global variables are usually used to define constants, if variables change based on the function call, we should define them in the function.

Maybe discussion is necessary before we go and create a PR for something you consider an issue.

Mec-iS:
> 1. Declare the keys as global

Functional programming shall avoid using globals

PR #351, opened 2019-02-21
Implement configuration module

Mec-iS:
The application's runtime procedure was really unclear. This fixes the duplication and configuration loading from environment variables


docker-compose.yml
```diff
@@ -5,31 +5,35 @@ services:
...
-      main:
-        aliases:
-          - postgres_db
+#  postgres:
+#    restart: always
+#    environment:
+#      - POSTGRES_USER=hydra
+#      - POSTGRES_PASSWORD=hailhydra
+#      - POSTGRES_DB=hydrus
+#    image: postgres:latest
```
xadahiya: We don't need this anymore.

Mec-iS: we will add that soon. the problem is that the tables creation script is unsafe.

hydrus/conf.py
```diff
@@ -0,0 +1,61 @@
...
+    DEBUG = bool(os.environ['DEBUG'])
+except KeyError:
+    DEBUG = False
+
+# load form environment (as many globals as possible shall be in
+# environment configuration)
+try:
+    PORT = int(os.environ['PORT'])
+    API_NAME = os.environ['API_NAME']
+    DB_URL = os.environ['DB_URL']
```
xadahiya: Don't you think there should be some default values in case env variables are not set or the user tries to run app.py without docker?

Mec-iS: these ones have to be set or the user has not control over the server.

I added the test variables in `travis.yml`

PR #354, opened 2019-02-23
Remove parser

Mec-iS:


requirements.txt
```diff
@@ -24,3 +24,4 @@ thespian==3.9.2
 Werkzeug==0.14.1
 PyYAML==4.2b1
 -e git+https://github.com/HTTP-APIs/hydra-python-core#egg=hydra_python_core
+-e git+https://github.com/HTTP-APIs/hydra-python-core#egg=hydra_python_core
```
xadahiya: This should be the parser repo

xadahiya:
One more thing, .travis.yml needs to be updated to include hydraspecs module in install or tests will fail

Mec-iS:
> One more thing, .travis.yml needs to be updated to include hydraspecs module in install or tests will fail

CI should install from `requirements.txt`

xadahiya:
For some reason `pip install -e` fails with Travis. Found this solution in stack overflow

Mec-iS:
`pip install -r requirements.txt` from the project directory?

xadahiya:
I guess anything with `-e` in requirements.txt will fail but only for travis. Locally it works fine

Mec-iS:
Probably we need to specify a version for the parser and core repos

xadahiya:
Yeah that might work

Mec-iS:
https://www.simonmweber.com/2014/11/23/git-pip-requirements.html

Mec-iS:
`pip` command fixed.

> Also the parser reference in cli.py needs to be updated.

Please open a issue for this

PR #362, opened 2019-03-16
Add `pip` and `setuptools` update to Dockerfile

Mec-iS:
Make the container to update `pip`


setup.py
```diff
@@ -16,10 +16,10 @@
...
 setup(name='hydrus',
       include_package_data=True,
-      version='0.1',
-      description='A space-based application for W3C HYDRA Draft',
-      author='W3C HYDRA development group',
-      author_email='public-hydra@w3.org',
+      version='0.2',
+      description='Hydra Ecosystem Flagship Server',
+      author='Hydra Ecosystem',
+      author_email='hydraecosystem@@googlegroups.com',
```
xadahiya: Is this okay? There are two @ 😅

Mec-iS: sorry, corrected

PR #369, opened 2019-03-21
Sync Master for patch 0.1.3

Mec-iS:


.travis.yml
```diff
@@ -10,3 +10,10 @@ install:
 script: python -m unittest discover
 env:
   - API_NAME=api PORT=8000
+deploy:
+  provider: pypi
+  user: tuned
+  password:
+    secure: vw9srAqCwa4QO7ITuyApU0Y/AazAfi6wWWXszuqa3/wAKPQ1fJ7T8JY/m3pKg0vMhoNp7hmI3v/nnyJPdTZGRGivhvsRUFC0EEC2c238DfPViAfsxEbHEpO45Ak6B3MR/fyvCmAZmMg3lXiwoP/EF5xvUNLu18Bnp1cdql5cd2iRyp6oTv/5/PWCq1a5phTpa3rP0rQvfqvmYERxgh4BjN9Uge2AFhUJqLvG1mGG/E0bWXGq2yOwI5orwcxFM0n2THsvXVD2+2fsM3oo6pXm74CgQJ7HCpnoJqssyXbc9f4aUnqUz1YK2V8PxE3WGv8jT9/IRwIy8ILIi5PQc9hsADDNxCABbxCsw3gPqJi+Gl99mmqwmo74Q9CypcwCE0paZEM5Ql37McRj91mPdpNB05FF4KzNM44eCIJOJbzTDst7MrvLBZqk2dGDtDMb5A9lO78PnCaJwEoRAeM6ktOtbd0ob/umwRTsoYzrJn4DXd4sbfQIGmnC7VdbCCfORCUHE2HjhRr38/LeC6YFbGnxmwNAIW+lavQC4WLpWwxJc8QQeIWQwuOGowwTcPs712bcdmV1Xaqf+nFA7RrAb/AvxxRlS8DZikbzEFQ5nU5UBd9V7gSN6501aq/xoTaslfuFeTwcwtNXOUdkmtZ5eIZHtyT5a1QjJ+jaDNHnwrh1snY=
```
de-sh: Just a question, but is it fine to have the password in here?

requirements.txt
```diff
@@ -22,6 +22,6 @@ six==1.11.0
 SQLAlchemy==1.2.10
 thespian==3.9.2
 Werkzeug==0.14.1
-PyYAML==4.2b1
+PyYAML
```
de-sh: If I am not wrong this just installs the latest version. Are we supposed to do that? We can have it set to the latest version of the package on which our repo works. i.e. v5.1

de-sh:
@Mec-iS so we upgraded to the latest version of PyYAML. But are we sure that there will not be an update to the package which might break our repo? [I would suggest we keep to v5.1](https://github.com/HTTP-APIs/hydrus/pull/369/files#r267789909)

requirements.txt
Mec-iS: this is fine as we automatically get all the security updates

chrizandr:
I also think we should add the latest stable version for PyYAML that is working instead of not specifying a version.

Mec-iS:
@vddesai1871 @xadahiya please confirm that this is good to merge

Mec-iS:
For the next time: please remember that `master` is not meant to contain `print` statements anywhere.

PR #371, opened 2019-03-22
Added test for cli.py

knakul853:
Fixes #

### Checklist

### Description

I added test for cli.py. I tested and it works fine :)

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,0 +1,127 @@
+"""Test for checking Cli"""
+import unittest
+import random
+import string
+import json
+import re
+import uuid
+from hydrus.app_factory import app_factory
+from hydrus.utils import set_session, set_doc, set_api_name
+from hydrus.data import doc_parse, crud
+from hydra_python_core import doc_maker
+from hydrus.samples import doc_writer_sample
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker, scoped_session
+from hydrus.data.db_models import Base
+import click
+from click.testing import CliRunner
+from cli import startserver
+
+def test_startserver():
+    runner = CliRunner()
+
+    #starting server with invalid params
+
+    result = runner.invoke(startserver,
+    ["--adduser","sqlite://not-valid","http://localhost",
+    "--port","serve"])
+
+    assert result.exit_code == 2
+
+
+     #starting ther server
+
+    result = runner.invoke(startserver,
+    ["--adduser","--api","--no-auth","sqlite:///:memory:",
+    "--hydradoc","--port","--no-token","http://localhost",
+    "serve"])  
+    assert result.exit_code == 0
+
+
+
+def gen_dummy_object(class_, doc):
+    """Create a dummy object based on the definitions in the API Doc."""
+    object_ = {
+        "@type": class_
+    }
+    if class_ in doc.parsed_classes:
+        for prop in doc.parsed_classes[class_]["class"].supportedProperty:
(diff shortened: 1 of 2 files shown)


de-sh:
@knakul853 seems like the tests succeed but aren't actually being run, maybe this is because of the absence of the __init__.py file. I am working on your submission and it seems like it doesn't even work with the addition of that file. I think we should be moving the tests folder as tests such as the pep8 one don't run on the files outside of the hydrus module(folder).

Also the Pull Request should be to branch `develop` instead of `master`.

tests/test_cli.py
```diff
@@ -0,0 +1,127 @@
...
+    @classmethod
+    def tearDownClass(self):
+        """Tear down temporary database and exit utilities"""
+        self.client.__exit__(None, None, None)
+        self.doc_util.__exit__(None, None, None)
+        self.session_util.__exit__(None, None, None)
+        self.api_name_util.__exit__(None, None, None)
+        self.session.close()
+
+    def setUp(self):
```
chrizandr: There are no assertions/checks here. How is this a test?

tests/test_cli.py
```diff
@@ -0,0 +1,127 @@
...
+from hydra_python_core import doc_maker
+from hydrus.samples import doc_writer_sample
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker, scoped_session
+from hydrus.data.db_models import Base
+import click
+from click.testing import CliRunner
+from cli import startserver
+
+def test_startserver():
```
chrizandr: This function is never called/used. What is it here for?

tests/test_cli.py
chrizandr: There are no assertions/checks here. How is this a test?

tests/test_cli.py
de-sh: He meant to make it a part of the test class but it's outside the scope. Would be wonderful if you could share a good resource to learn testing as we are new to it.

knakul853: @chrizandr  yeah i forget to put in test class now i changed ,will make new PR

tests/test_cli.py
knakul853: @chrizandr  what assertion i can make hear i saw another test files that does not have any aasertions for this .

tests/test_cli.py
de-sh: @knakul853 don't make a new PR, you can always undo a commit and force push the changes.

ShravanDoda: @knakul853 kindly take a look at the docs
unittest -> https://docs.python.org/3/library/unittest.html
unittest.mock -> https://docs.python.org/3/library/unittest.mock.html

tests/test_cli.py
de-sh: I will make a PR to your fork with the changes I think are worth it. Let's collaborate on it

Meanwhile can someone suggest a good resource to learn testing? :sweat_smile:

ShravanDoda: I think python docs for testing are pretty good

tests/test_cli.py
knakul853: @ShravanDoda  thanks

tests/test_cli.py
knakul853: @de-sh yep

knakul853:
I am closing this pr and opening a new one because by mistake I opened the pull request in comparison with master branch. sorry for inconvenience.

Issue #377, opened 2019-03-25
Add tests for retrieving Items from collections


vddesai1871:
### I'm submitting a 
- [x] bug report.
- [x] feature request.

### Current Behaviour:

~We have tests for checking operations over individual members of collection, but we do not have any test to check response for collection (for example checking response for api/XyzCollection).~
Add test to check `member` field of collection response holds valid items/elements. 
### Expected Behaviour:

We need a simple test to check status code returned by dereferencing a member element.
### Steps to reproduce:

In crud.py line 559  and 563 sets the uri of member elements. 
```
"@id": "/{}/{}/{}".format(API_NAME, path, instance_.id),
```
change it to anything else like
```
"@id": "/{}/xyz/".format(API_NAME, path, instance_.id),
````
every tests will pass as it is not being tested in any of the tests.
### Do you want to work on this issue?

yes

chrizandr:
We do have `test_Collections_GET` and `test_Collections_PUT` in `test_app.py`.

Is there anything I am missing? I added those tests a long time ago in #48 . We are already checking responses for every collection, what seems to be missing there?

vddesai1871:
Yes the test I mentioned in the issue is already in test_Collection_GET, I do not remember how I missed it. Still I'll reevaluate it and then close it if I find nothing.

vddesai1871:
I think it was for checking the validity of member item URIs in the collection response. I encountered it while working on #378 , but yes the description I gave was wrong, I have updated it and I am creating a PR for it.

chrizandr:
Merged in #398

PR #378, opened 2019-03-25
remove use of finalization on collection response

vddesai1871:
Related to #377
In #326 I added finalize function to finalize response body of objects, it should not be used on collections, but it is being used on line 244 of resources.py (It was added by me during resolving conflicts), As there were no tests to check collection responses, this bug was merged in develop, I have created an issue for adding tests here #377.
This PR removes that line.

### Checklist


vddesai1871:
@Mec-iS @xadahiya @chrizandr Can you have a look and merge it? It's an important one.

PR #380, opened 2019-03-26
Add long description to setup.py

Mec-iS:


setup.py
```diff
@@ -18,6 +18,7 @@
       include_package_data=True,
       version='0.2.3',
       description='Hydra Ecosystem Flagship Server. Deploy REST data for Web 3.0',
+      long_description=open("README.md").read(),
```
chrizandr: Wouldn't this read the Markdown formatting text too?

Mec-iS: yea, md is supported by that keyword

Issue #349, opened 2019-02-02
Implement migrations with Alembic


Mec-iS:
### I'm submitting a 
- [x] feature request.

### Current Behaviour:
The server should provide a standard migration tools to handle any change in schemas. The most popular migration tool for SQLAlchemy is [Alembic](https://alembic.sqlalchemy.org/en/latest/)

### Expected Behaviour:
All changes in the database schema should be tracked and reproducible

aribalam:
@Mec-iS I am not able to understand the goal of this feature. What are we going to achieve with this. Is it related with including property attributes in the database (#226)?
Currently only the relationships between classes and properties (the triples) are being stored in a database. With property constraints and other sorts of validation to be performed by `HydraDoc` in `resources.py` as was decided in #226, there would hardly be any change in schemas.

chrizandr:
@aribalam There might be changes we make to the DB in that would change the schema. It is a good practice to have migrations in place to make the transition easier, without having to clean the entire database.

Mec-iS:
Implementing Alembic is a step-stone for allowing on-the-fly ApiDoc updates (e.g. change the classes definition without hard-coding the client, or "dynamic paths"). A note:

> For now we want the mechanism to run smoothly, leaving all these concerns for the second phase of the implementation. So probably at the moment we want to apply the migrations for the new models and restart the server. In a second phase we will carefully look into the data consistency issue (the first solution that comes to my mind is to have a temporary cache for all the operations that happen at server restart or to keep an log of events so to re-establish consistency after the new models are up)

Guttz:
Okay so while implementing this I realized that **there are two situations into the same issue**, which also had me confused for a while now:

**1 -** Alembic migrations with the intent of **on-the-fly ApiDoc**(Modification on the structure of the API)
**2 -** Migrations with the intent of occasionally giving the users a way to maintain their data **in the case we change how we store the relationships** inside hydrus

##### On-the-fly ApiDoc updates
And the thing is, with the way hydrus DB structure was implemented, On-the-fly ApiDoc updates are already supported as you can see in the demo gif below. You can edit the ApiDoc, modify properties, delete classes and by the fact that the database is used only to map the relations, hydrus will be able to persist the data. 
![Dynamic Endpoint Gif](https://i.imgur.com/tatfapU.gif)
(Click on the gif to expand) 

##### DB Schema modification

As described by @chrizandr:
> @aribalam There might be changes we make to the DB in the future that would change the schema...

[...]

PR #381, opened 2019-03-28
Adding migrations with Alembic

HarsheetKakar:
Fixes Implement migrations with Alembic #349

### Checklist

### Description
Alembic database migration tool added for future migrations in database
### Change logs

- Alembic in requirements.txt
- set up for Alembic for single database


.gitignore
```diff
@@ -110,5 +110,5 @@ ENV/
 # Keys
 *.pem
 
-# Pipenv generated
-Pipfile*
\ No newline at end of file
+# Pipenv generated
```
de-sh: Undo changes to this file

de-sh:
@HarsheetKakar the PEP8 tests say that the following issues exist in the file `hydrus/data/db_models.py`
```
14:89: E261 at least two spaces before inline comment
14:90: E262 inline comment should start with '# '
14:101: E501 line too long (131 > 100 characters)
```

hydrus/data/db_models.py
```diff
@@ -8,9 +8,12 @@
 from typing import Any
 import datetime
 import uuid
+import os
 # from hydrus.settings import DB_URL
 
-engine = create_engine('sqlite:///database.db')
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.realpath(__file__)))) #to put database file in the rootdirectory
```
de-sh: This is where all issues occur. Our PEP8 test fails at 101 chars per line. Place the comment on top of the code.

HarsheetKakar: done, thanks for pointing out

Mec-iS:
we need documentation on how to run migrations

HarsheetKakar:
> we need documentation on how to run migrations

please check if migrations.md works for our purposes

migrations.md
```diff
@@ -3,24 +3,30 @@
...
-config.set_main_option('sqlalchemy.url', 'sqlite:///{}/database.db'.format(DATABASE_DIR))
-'''
+
+    DATABASE_DIR = <location of database.db>
+    config.set_main_option('sqlalchemy.url','sqlite:///{}/database.db'.format(DATABASE_DIR))
+
+
+
 
 Now you can make changes in db_model.
```
Guttz: Currently hydrus creates db_model with the assistance of doc_parse based on the Hydra API provided by the user.  That said, I don't think that's quite the use case we would have in the Hydra Ecosystem. I'll comment below how I personally think that could be done.

Guttz:
So as a useful feature to be added for hydrus I can see the migration provided in two ways:

1. Instruct the user to manually run hydrus once(which means creating a copy of the db, update Hydra Doc and run hydrus) and so with the updated db_models then run alembic migrate with the old copy of the db.. Which would be basically using alembic by itself

2. Create a hydrus option (hydrus --migrate) that calls a script that does the above: update db_models based on Hydra Doc, run it with the database and outputs the upgrade script - a bonus nice addition would to potentially try to run the migration automatically. In my mind a nice flow for the user would be the following: 

image

HarsheetKakar:
> So as a useful feature to be added for hydrus I can see the migration provided in two ways:
> 
> 1. Instruct the user to manually run hydrus once(which means creating a copy of the db, update Hydra Doc and run hydrus) and so with the updated db_models then run alembic migrate with the old copy of the db.. Which would be basically using alembic by itself
> 2. Create a hydrus option (hydrus --migrate) that calls a script that does the above: update db_models based on Hydra Doc, run it with the database and outputs the upgrade script - a bonus nice addition would to potentially try to run the migration automatically. In my mind a nice flow for the user would be the following:
> 
> image

1. can be done, I can just add it above env.py part
2. for this we need someone who knows click, im currently not working with it

Mec-iS:
@Guttz please pick this up and provide a complete patch, You can just fork and open a new PR.

Guttz:
New PR about migrations at https://github.com/HTTP-APIs/hydrus/pull/389.

Issue #304, opened 2018-12-08
Mechanism for pagination


PrajwalM2212:
### I'm submitting a 
- [ ] feature request.

### Expected Behaviour:
Hydra should allow  to retrieve collections by pages so that I can show first batch of items as soon as possible. This is also discussed in https://github.com/HydraCG/Specifications/blob/master/drafts/use-cases/3.2.pagination.md.

Mec-iS:
Yes, pagination is a very important feature.

sreetamparida:
I want to work on this issue

HarsheetKakar:
https://github.com/djrobstep/sqlakeyset came across this we can use parts of the code or put it in requirements... give your feedback.

PR #385, opened 2019-04-23
Pagination

HarsheetKakar:
solves #304

### Checklist
- [x] My branch is up-to-date with upstream/master branch.

### Description
pagination mechanism for hydrus using sqlalchemy-paginator

### Change logs

Create paginator function added
pagewise get function added works like get collection but in paginated form


Mec-iS:
You should add more test cases.

@xadahiya @chrizandr please take a look

hydrus/data/crud.py
```diff
@@ -47,6 +47,8 @@
 # from sqlalchemy.orm.session import Session
 from sqlalchemy.orm.scoping import scoped_session
 from typing import Dict, Optional, Any, List
+from typing import Dict, Optional, Any, List, Union, Iterator
```
xadahiya: Duplicate imports

requirements.txt
```diff
@@ -19,9 +19,10 @@ pyparsing==2.2.0
...
 pytz==2018.5
 six==1.11.0
-SQLAlchemy==1.2.10
+SQLAlchemy
 thespian==3.9.2
-Werkzeug==0.14.1
-PyYAML==5.1
+Werkzeug
+PyYAML
+git+git://github.com/HarsheetKakar/sqlalchemy-paginator.git
```
xadahiya: Is this a good idea to use a package that was last updated in 2015?

@chrizandr @Mec-iS

Mec-iS: we don't want non-ecosystem/external dependencies that are not constantly updated and/or on PyPi

HarsheetKakar: its a very small mistake in their code, should i copy the whole code in hydrus itself? Its a small library

HarsheetKakar:
> You should add more test cases.
> 
> @xadahiya @chrizandr please take a look

sure, will continue when exams are over/or when free in between

requirements.txt
xadahiya: Doesn't matter if it's a small library or not. If there's a mistake in the main library that clearly indicates that it might not be written very well. Also, it hasn't been updated since 2015.

PR #389, opened 2019-05-20
Setting up migrations with Alembic

Guttz:
Fixes https://github.com/HTTP-APIs/hydrus/issues/349 and closes https://github.com/HTTP-APIs/hydrus/pull/381

### Checklist

### Description

Providing a pre-configured extension of Alembic to facilitate hydrus DB Migrations. Documentation available on how to run/write a migration script and also how to use an Alembic auto migration generator, with the proper instructions.  
 
#### Changed
* Pre-configured Alembic for hydrus
* Basic demo migration flow 
* Auto Generating Migrations

#### Fixed

* hydrus user exception when restarting hydrus with same DB

@Mec-iS @chrizandr Take a look at it


hydrus/data/alembic.ini
```diff
@@ -0,0 +1,75 @@
...
+# to alembic/versions.  When using multiple version
+# directories, initial revisions must be specified with --version-path
+# version_locations = %(here)s/bar %(here)s/bat alembic/versions
+
+# the output encoding used when revision files
+# are written from script.py.mako
+# output_encoding = utf-8
+
+# hydrus users please change this line with the address to your Database
+sqlalchemy.url = sqlite:///database.db
```
chrizandr: Can we change this to take the URL from `os.environ`? The same as we did in `conf.py`.

This might be helpful: https://stackoverflow.com/questions/37890284/ini-file-load-environment-variable

hydrus/data/migrations/README.md
```diff
@@ -0,0 +1,90 @@
...
+-----------------
+First make sure you set up hydrus as the [instructions here](https://github.com/HTTP-APIs/hydrus/blob/master/README.md).
+
+## Basic demo migration flow 
+
+With the default hydrus clone run it with a sqlite database:
+
+    hydrus serve --dburl sqlite:///hydrus/data/database.db
+
+#### Point to your DB Url(only if not using sqlite):
```
chrizandr: Even when using sqlite, the DB name is a variable, so this text needs to be changed.

Guttz: Yes, I guess that's a bit confusing. I'll make a change.

hydrus/data/alembic.ini
Guttz: Sure, I will take a look. I can also just import DB_URL from conf.py, makes sense?

Guttz: Sure, I will take a look. I can also just import DB_URL from conf.py.

Mec-iS:
Maybe we can consider creating a Jinja template for the INI file, so we can run a command that generates it by taking the variables from existing modules.

Guttz:
Took a look at Jinja and looks interesting, but to be honest with our current "simple one-Database" structure the only variable that had to be set at Alembic.ini was the sqlalchemy.url, that as suggested above by @chrizandr, was already ported and is being dynamically set now. So I guess it's a good note for the future extensions of the migration but as for now we don't need it yet.

hydrus/data/alembic.ini
Guttz: It's done, it works life conf.py and also rewrote the readme.

hydrus/data/migrations/README.md
```diff
@@ -0,0 +1,92 @@
...
+After changing the db schema at hydrus/data/db_models.py accordingly to your wish(and running *python setup.py install* again to update your build), run:
+
+	alembic revision --autogenerate -m "autogenerated changes" 
+
+You should get something *like hydrus/data/migrations/versions/0bca3dcf0826_autogenerated_changes.py* ... done. You should head over to this file and double-check/compliment your migration script. After this, just run your migration script. 
+
+    alembic upgrade head
+
+
+#### IMPORTANT - Quotting Alembic official page:
```
chrizandr: Typo in the word "quoting"

PR #391, opened 2019-05-29
Add mechanism for pagination

### LINKED ISSUE #304: Mechanism for pagination

vddesai1871:
Fixes #304  

### Checklist

### Description

Adds various functions to provide a mechanism for pagination.

#### Additions/changes
* CLI arguments to enable/disable pagination and set page size, and utility functions related to them.
* Changes in `get_collection` method in crud.py and get method of `ItemCollection` class in resources.py to implement actual pagination functionality.
* Exception class to handle excpetions related to invalid page number/id.
* Tests for pagination.

### DIFF
--- a/hydrus/data/crud.py
+++ b/hydrus/data/crud.py
@@ -43,7 +43,8 @@ from hydrus.data.exceptions import (
     PropertyNotFound,
     NotInstanceProperty,
     NotAbstractProperty,
-    InstanceNotFound)
+    InstanceNotFound,
+    PageNotFound)
 # from sqlalchemy.orm.session import Session
 from sqlalchemy.orm.scoping import scoped_session
 from typing import Dict, Optional, Any, List
@@ -584,6 +585,9 @@ def get_collection(API_NAME: str,
         last = total_items // page_size
     else:
         last = total_items // page_size + 1
+
+    if page < 1 or page > last:
+        raise PageNotFound(str(page))
     collection_template["view"] = {
         "@id": "/{}/{}?page={}".format(API_NAME, path, page),
         "@type": "PartialCollectionView",
--- a/hydrus/data/crud.py
+++ b/hydrus/data/crud.py
@@ -516,8 +516,8 @@ def get_collection(API_NAME: str,
                    type_: str,
                    session: scoped_session,
                    paginate: bool,
-                   page_size: int = None,
-                   page: int = None,
+                   page_size: int,
+                   page: int = 1,
                    path: str = None) -> Dict[str, Any]:
     """Retrieve a type of collection from the database.
     :param API_NAME: api name specified while starting server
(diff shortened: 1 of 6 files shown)


hydrus/resources.py
```diff
@@ -226,6 +232,8 @@ def get(self, path: str) -> Response:
         """
         Retrieve a collection of items from the database.
         """
+        page = request.args.get('page', 1)
+        print(type(page))
```
xadahiya: Remove redundant print statements

hydrus/data/crud.py
```diff
@@ -514,17 +515,22 @@ def update(id_: str,
 def get_collection(API_NAME: str,
                    type_: str,
                    session: scoped_session,
-                   path: str = None) -> Dict[str,
-                                             Any]:
+                   paginate: bool,
+                   page_size: int = None,
+                   page: int = None,
```
xadahiya: Instead of setting the page's default value to 1 in resources.py L235, I think you should do that here. That way crud.py will be a little less dependent on `resources.py`. Don't you think?

vddesai1871: Yes makes sense but the current way avoids an extra check. So if we assign the default value in the function definition of crud.py, I will have to make two statements in resources.py to call get_collection function. 
```
If page == None:
     get_collection(.....) # not providing value for page parameter
else:
     get_collection(...,page=page,....) 
```
Why is it so?
Because if I only use the current statement `get_collection(....,page=page,...)` the `page` value is always being passed(None in the case of missing query parameter), so the default value assignment never happens in the function as a value is always provided.

1) So to set the default value in crud.py one way is to add an extra check as shown in the code snippet above.
2) Manually set the default value in crud.py. 
```
page == None:
     page = 1
```
Yes, it acts differently than setting value in the argument list by ` page: int = 1` because here default value is applied only when page parameter is not passed, if page == None is passed then the default value is not set.

So to set the default value in crud.py I would have to do an extra check(option 1) OR set the default value in the function body of get_collection instead of argument list (option 2).

hydrus/data/crud.py
```diff
@@ -563,6 +574,37 @@ def get_collection(API_NAME: str,
             object_template = {
                 "@id": "/{}/{}Collection/{}".format(API_NAME, type_, instance_.id), "@type": type_}
         collection_template["members"].append(object_template)
+
+    # If pagination is disabled then stop and return the collection template
+    if not paginate:
```
Mec-iS: all checks should be explicit by using `is`: i.e. `if var is None` or `if var is False` etc. this improves understandability.

hydrus/data/crud.py
xadahiya: I think option 1 would be a better choice here. I'm saying so because
- It'll make our codebase more explicit i.e if page is None then we call `get_collection` without the page argument and crud.py takes care of the defaults, if we have page value then we pass page value as an argument and the default value is overridden.
- It'll also solve the issue I initially pointed out of making `crud.py` less dependent on `resources.py`

I don't think adding an extra check would hurt us in any way. Thoughts?

vddesai1871: I agree, I just wanted to show all the options we have so I can get those reviewed and make a better decision. **Going with option 1**

PR #393, opened 2019-05-29
Merge from develop

Mec-iS:


Mec-iS:
Please double-check and eventually give :+1:

PR #400, opened 2019-06-05
Adding DELETE operation to drone collection

Guttz:
Fixes #

### Checklist

### Description
Adding DELETE to sample Drone API spec

@xadahiya was it just not necessary before or was there a specific reason to not have it?

### DIFF
--- a/examples/drones/doc.py
+++ b/examples/drones/doc.py
@@ -423,6 +423,23 @@ doc = {
                     ],
                     "returns": "vocab:Drone",
                     "title": "GetDrone"
+                },
+                {
+                    "@type": "http://schema.org/DeleteAction",
+                    "expects": "null",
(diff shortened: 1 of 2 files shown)


Mec-iS:
are you adding this to the resource or the collection? it is useful to delete single resources but not much deleting an entire collection

Guttz:
It's directly on the resource, I've been using the Drone Collection for the client documentation so I'll try to maintain it updated as we move on.

Mec-iS:
all changes to the Flock vocabulary have to be made in https://github.com/HTTP-APIs/hydra-flock-vocab the example in hydrus should be pulled from there.

Guttz:
Ok, I'll be updating there

xadahiya:
@Guttz this was not required for the demo so we didn't use it.

Issue #392, opened 2019-05-29
Change response message structure


vddesai1871:
### I'm submitting a 
- [ ] bug report.
- [x] feature request.

### Current Behaviour:

Right now hydrus returns a string as response message which contains the information related to object's ID. If the client wants to access that, it has to use a regular expression to extract the ID from the message.
### Expected Behaviour:

We can return a response in which object ID is separated from the verbal message, so a client can access it directly.
Ex. 
```
{
"message": "Object successfully created",
"object_id": "deede-w9d2wo-dkwd-fefrr3rt4"
}
```

@Guttz anything you want to add?
### Do you want to work on this issue?

YES

Guttz:
That's what I think is important for now. I guess a third value as the timestamp may turn out to be necessary in the future for sync conflicts, but if it eventually appears as crucial we do it. Thanks, @vddesai1871!

Mec-iS:
also adding a boolean in a `error` key if the status code is not 2xx or 3xx would be nice to have

chrizandr:
Does the spec say anything about the format of response messages? We should find that out. 

Alternatively, we can return the modified object in case of PUT/POST. For DELETE the status code should be sufficient. The reason being that we would have a more standard response rather than ambiguous messages that we need to parse or custom objects where we define our own fields.

Also since all responses are HTTP responses, they have status codes by default. Unless there is an error we should use code 200 for responses. We already have appropriate status codes based on the error implemented in the exceptions.

vddesai1871:
Yes following the spec would be the best option to go.
As I mentioned in the issue the issue was created to separate the object ID from the textual message. 
So according to spec(the portion you linked in the other issue), we can directly return a `Status` or `Error` object in the response. And looking at the example given there and looking at core.jsonld (the actual vocab) the response will look something like
```
{
 "@context": "../../abc.jsonld",
 "@type": "Status",
 "title": "Instance creation successful",
 "description": "Instance with ID X successfully created."
}
```
It's certainly better but how can I separate the ID from the description message? It will probably involve adding an extra non-standard field. If I am not wrong the only time a client can get the benefit of separated ID is when it creates a new object. So should we add the extra field for object_id or just use the current response mechanism adjusting it to the spec format?

chrizandr:
The spec mentions subclassing the Error and Status classes. It also allows us to add additional identifiers to the messages. 
> Furthermore, a Status or Error returned by the server can also be given an identifier.
[...]

PR #402, opened 2019-06-16
Use HydraStatus and HydraError to return status response

vddesai1871:
Fixes #

### Checklist

### Description

Returns standard response status body with help of Hydra Status and Hydra Error. Also uses them in the sample documents.

## Key Design decisions
1. Avoids the use of `@id` to separate id from the message. Looking at the Spec `@id` field is supposed to point towards a detailed version of the same Status message and not towards the object on which the operation had been done.(http://www.hydra-cg.com/spec/latest/core/#description-of-http-status-codes-and-errors) So setting it to the object id would be incorrect. (We have already resolved the issue about separating id by using `Location` header value at hydra-agent to get the id of the newly created object(#392 ).
2. Linking of possibleStatus defined in ApiDoc to the statuses returned by hydrus: I tried to do this by linking this via statusCode(But it was not possible for all scenarios as a single HTTP status code might be used for more than one errors/statuses.) So I can't think of a way to link possibleStatus with statuses returned by hydrus. If you have any other appraoch in mind you can write here or create a new issue. (Right now we are going with just using standard Spec format to return statuses and errors without linking the possibleStatus defined in ApiDoc to the statuses we return.

### DIFF
--- a/hydrus/auth.py
+++ b/hydrus/auth.py
@@ -1,6 +1,7 @@
 from typing import Union
 
 from flask import jsonify, Response, request
+from hydra_python_core.doc_writer import HydraError
 from hydrus.utils import get_session, get_token, get_authentication
 
 from hydrus.data.user import (create_nonce, check_authorization,
@@ -48,8 +49,8 @@ def verify_user() -> Union[Response, None]:
             token = add_token(request, get_session())
             return token_response(token)
     except Exception as e:
-        status_code, message = e.get_HTTP()  # type: ignore
-        return set_response_headers(jsonify(message), status_code=status_code)
+        error = e.get_HTTP()  # type: HydraError
+        return set_response_headers(jsonify(error.generate()), status_code=error.code)
     return None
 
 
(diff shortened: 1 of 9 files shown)


vddesai1871:
> I think it is important that we return the ID of the object in the response rather than in the headers.

I agree.
Subclassing will do the work.
Looking at the body of different error and status messages, there are three fields which we might want to have in these subclasses (id, type, and object_list(used for operations involving multiple items))
We can directly subclass HydraStatus to HydrusStatus and HydraError to HydrusError. But a better approach would be to subclass HydraStatus to get HydrusStatus and then subclass HydrusStatus to get HydrusError. Besides that the question is where should I put those classes? We can add them at `hyrda-python-core` or add them at hydrus maybe in a new file in the data directory. I prefer the latter.

chrizandr:
It would be better to keep the spec separate from hydrus specific code. Let's implement these in hydrus rather than in core since these classes are not part of the spec. 

Also remember since we are defining our own classes, we need to add them to the API doc

PR #406, opened 2019-07-08
Add mechanism to retrieve nested collections

### LINKED ISSUE #386: hydrus can't show nested collections
### I'm submitting a 
- [x] bug report.
- [x] feature request.

### Current Behaviour:

hydrus can't show nested collections. (eg. collection of comments on a particular issue/article)
### Expected Behaviour:

Quoting @Mec-iS 
>we possibly want to return a hyperlink to the collection, not the full collection. same we did for nested object in Flock-Demo.

#### Possible approach
 we can have a hyperlink to the endpoint of nested collection
something like
```
{ 
"@type": "Article", 
"@id": "api/articles/id", 
... 
"comments": "api/articles/id/comments", 
... 
} 
```

### Steps to reproduce:

Run hydrus with real-world-app API Doc I submitted at https://github.com/HTTP-APIs/hydrus-real-world-app/pull/2 (The API Doc is not complete but can be used to reproduce this issue by adding an article and trying to get/retrieve comments related to that article)

Yes

vddesai1871:
Fixes #386 

### Checklist

### Description

Uses the approach suggested by @xadahiya at #386 to store a stringified list of children ids and internally uses  parent_id to update those lists.

There are a couple of issues which needs to be discussed.
1) **Identifying parent-child relationships**
When a new item is added to the child collection(`CommentCollection`) we need to identify parent class, which is quite easily done by searching for a class which has `vocab:CommentCollection` as `property`. But it is hard to identify the field which contains the `id` of the parent class. I think we can't simply use `schema.org/identifier` as it might be used to describe any other kind of id. 
To tackle it temporarily so I can move ahead with other parts of implementation I am having the `property` predicate of ParentURI supportedProperty set to the parent class(for example Comment class's issue_id field will have `vocab:Issue` as property. But this parent_id/uri field might only contain an id or a URI, but by property value `vocab:Issue` it seems the whole object is being sent when creating new Comment(As `State` object is contained in the `Drone` object while creation.) Line 322-355 in resources.py uses this to identify field which contains parent_id. It works well but it is not semantically correct. If ApiDoc maker always uses schema.org/parentItem or some other value as `property` value of parent_id field then I can avoid what I am doing right now and can easily identify which field is the `parent_id` field.(crud_test is failing while creating dummy_objects because of these new additions in the sample doc, I'll adapt tests when we some concrete solution to this problem.)
So basically the question is *how hydrus finds out which field of received objects contains parent_id?*
2) **Current sample ApiDoc does not have any such parent-nestedCollection realtions**
To test the mechanism I added a new field in Drone class which has `StateCollection` as property. We might want to have some other more meaningful way to add such a relationship or have another sample ApiDoc with such a relationship.

Other straight-forward tasks to be done:
1. Adding pagination and serch over nestedCollection.
2. Adding tests.

### DIFF
--- a/hydrus/app_factory.py
+++ b/hydrus/app_factory.py
@@ -3,7 +3,7 @@ from flask_cors import CORS
 from flask_restful import Api
 
 from hydrus.resources import (Index, Vocab, Contexts, Entrypoint,
-                              ItemCollection, Item, Items)
+                              ItemCollection, NestedCollection, Item, Items)
 
 
 def app_factory(api_name: str = "api") -> Flask:
@@ -38,5 +38,9 @@ def app_factory(api_name: str = "api") -> Flask:
         "/{}/<string:path>/add/<int_list>".format(api_name),
         "/{}/<string:path>/add".format(api_name),
         "/{}/<string:path>/delete/<int_list>".format(api_name))
+    api.add_resource(
+        NestedCollection,
+        "/{}/<string:collection_type>/<uuid:id_>/<string:nested_collection_type>".format(api_name),
+        endpoint="nested_collection")
 
     return app
--- a/hydrus/data/crud.py
+++ b/hydrus/data/crud.py
@@ -130,11 +130,13 @@ def get(id_: str, type_: str, api_name: str, session: scoped_session,
     return object_template
 
 
-def insert(object_: Dict[str, Any], session: scoped_session,
+def insert(object_: Dict[str, Any], session: scoped_session, parent_id: str = None, parent_type: str = None,
            id_: Optional[str] = None) -> str:
     """Insert an object to database [POST] and returns the inserted object.
     :param object_: object to be inserted
     :param session: sqlalchemy scoped session
+    :param parent_id: id of the parent object
+    :param parent_type: type of the parent object
     :param id_: id of the object to be inserted (optional param)
     :return: ID of object inserted
 
@@ -166,6 +168,17 @@ def insert(object_: Dict[str, Any], session: scoped_session,
         instance = Instance(type_=rdf_class.id)
     session.add(instance)
     session.flush()
+    if parent_id is not None:
+        try:
+            parent_instance = session.query(Instance).filter(
+                        Instance.id == parent_id).one()
+            if parent_instance.children is None:
+                parent_instance.children = str(instance.id)
+            else:
+                parent_instance.children += ",{}".format(str(instance.id))
+        except NoResultFound:
+            session.close()
+            raise InstanceNotFound(parent_type, parent_id)
 
     for prop_name in object_:
 
@@ -616,6 +629,46 @@ def get_collection(API_NAME: str,
         collection_template["view"]["previous"] = "{}page={}".format(recreated_iri, page-1)
     if page != last:
         collection_template["view"]["next"] = "{}page={}".format(recreated_iri, page + 1)
+    return collection_template
+
+
+def get_nested_collection(api_name: str, parent_collection_name:str, parent_object_type: str,
+                          parent_id: str, nested_collection_name: str,
+                          object_type: str, session: scoped_session) -> Dict[str, Any]:
+    """Get items of a nested collection.
+    :param api_name: api name specified while starting server.
+    :param parent_collection_name: name of the parent collection.
+    :param parent_object_type: type of the parent object.
+    :param parent_id: id of the parent object.
+    :param nested_collection_name: name of the nested collections.
+    :param object_type: type of the objects of the nested collection.
+    :param session: sqlalchemy scoped session.
+    :return: response containing a nested collection.
+    """
+    collection_template = {
+        "@id": "/{}/{}/{}/{}".format(api_name, parent_collection_name, parent_id, nested_collection_name),
+        "@context": None,
+        "@type": nested_collection_name,
+        "members": list()
+    }  # type: Dict[str, Any]
+
+    try:
+        parent_instance = session.query(Instance).filter(
+                    Instance.id == str(parent_id)).one()
+        if parent_instance.children is None:
+            object_list = []
+        else:
+            object_list = parent_instance.children.split(',')
+    except NoResultFound:
+        session.close()
+        raise InstanceNotFound(parent_object_type, parent_id)
+
+    for item in object_list:
+        object_template = {
+            "@id": "/{}/{}/{}".format(api_name, nested_collection_name, item),
+            "@type": object_type
+        }
+        collection_template["members"].append(object_template)
     return collection_template
 
 
--- a/hydrus/data/db_models.py
+++ b/hydrus/data/db_models.py
@@ -53,6 +53,7 @@ class Instance(Base):
     type_ = Column(String, ForeignKey("classes.id"), nullable=True)
     created = Column('created', DateTime, default=func.now())
     last_modified = Column('last_modified', DateTime, onupdate=func.now())
+    children = Column(String, default=None)
 
 
 class BaseProperty(Base):
--- a/hydrus/helpers.py
+++ b/hydrus/helpers.py
@@ -166,6 +166,9 @@ def finalize_response(class_type: str, obj: Dict[str, Any]) -> Dict[str, Any]:
             obj.pop(prop.title, None)
         elif 'vocab:' in prop.prop:
             prop_class = prop.prop.replace("vocab:", "")
+            if prop_class in get_doc().collections:
+                obj[prop.title] = obj["@id"] + "/" + prop_class
+                continue
             nested_path, is_collection = get_nested_class_path(prop_class)
             if is_collection:
                 id = obj[prop.title]
@@ -202,6 +205,8 @@ def generate_iri_mappings(class_type: str, template: str, skip_nested: bool =Fal
     ).parsed_classes[class_type]["class"].supportedProperty:
         if "vocab:" in supportedProp.prop and skip_nested is False:
             prop_class = supportedProp.prop.replace("vocab:", "")
+            if prop_class in get_doc().collections:
+                continue
             template, template_mapping, skip_delimiter = generate_iri_mappings(prop_class, template, skip_nested=True,
                                                                                skip_delimiter=skip_delimiter,
                                                                                parent_prop_name=supportedProp.title,
--- a/hydrus/resources.py
+++ b/hydrus/resources.py
@@ -316,9 +316,34 @@ class ItemCollection(Resource):
                         obj_type, object_):
                     # If Item in request's JSON is a valid object ie. @type is a key in object_
                     # and the right Item type is being added to the collection
+                    parent_id = None
+                    parent_type = None
+                    # Check if the new object is a child of any other object and extract parent id
+                    for supportedProp in get_doc().parsed_classes[obj_type]['class'].supportedProperty:
+                        if "vocab:" in supportedProp.prop:
+                            parent_class = supportedProp.prop.replace("vocab:", "")
+                            if parent_class in get_doc().collections:
+                                continue
+                            for property_of_parent in get_doc().parsed_classes[parent_class]['class'].supportedProperty:
+                                if "vocab:" in property_of_parent.prop:
+                                    nested_collection_of_parent = property_of_parent.prop.replace("vocab:", "")
+                                    if nested_collection_of_parent == path:
+                                        parent_id = object_[supportedProp.title].split('/')[-1]
+                                        parent_type = parent_class
+                                        break
+                        if parent_id is not None:
+                            break
                     try:
-                        # Insert object and return location in Header
-                        object_id = crud.insert(object_=object_, session=get_session())
+                        if parent_id is not None:
+                            # Insert object and return location in Header
+                            print(parent_id)
+                            print(parent_type)
+                            object_id = crud.insert(object_=object_, parent_id=parent_id,
+                                                    parent_type=parent_type,session=get_session())
+                        else:
+                            # Insert object and return location in Header
+                            object_id = crud.insert(object_=object_, session=get_session())
+
                         headers_ = [
                             {"Location": "{}{}/{}/{}".format(
                                 get_hydrus_server_url(), get_api_name(), path, object_id)}]
@@ -520,6 +545,45 @@ class Items(Resource):
                 return set_response_headers(jsonify(error.generate()), status_code=error.code)
 
         abort(405)
+
+
+class NestedCollection(Resource):
+    """Handle operation related to NestedCollection."""
+
+    def get(self, collection_type:str, id_: str, nested_collection_type: str) -> Response:
+        """
+        Retrieve a collection of items from the database.
+        """
+        auth_response = check_authentication_response()
+        if isinstance(auth_response, Response):
+            return auth_response
+        endpoint_ = checkEndpoint("GET", collection_type)
+        if not endpoint_['method']:
+            # If endpoint and Get method not supported in the API
+            abort(endpoint_['status'])
+        elif collection_type in get_doc().collections:
+            # If endpoint and GET method is supported in the API
+            # and collection name in document's collections
+            nested_collection = get_doc().collections[nested_collection_type]["collection"]
+
+            # title of HydraClass object corresponding to the nested collection
+            nested_obj_type = nested_collection.class_.title
+
+            parent_collection = get_doc().collections[collection_type]["collection"]
+            parent_obj_type = parent_collection.class_.title
+
+            try:
+                # Get nested collection details from the database
+                response = crud.get_nested_collection(
+                    get_api_name(), parent_collection_name=collection_type, parent_object_type=parent_obj_type,
+                    parent_id=id_, nested_collection_name=nested_collection_type,
+                    object_type= nested_obj_type, session=get_session())
+                return set_response_headers(jsonify(hydrafy(response, path=nested_collection_type)))
+
+            except (ClassNotFound, PageNotFound, InvalidSearchParameter) as e:
+                error = e.get_HTTP()
+                return set_response_headers(jsonify(error.generate()), status_code=error.code)
+
 
 
 class Contexts(Resource):
--- a/hydrus/samples/hydra_doc_sample.py
+++ b/hydrus/samples/hydra_doc_sample.py
@@ -119,10 +119,10 @@ doc = {
                 },
                 {
                     "@type": "SupportedProperty",
-                    "property": "http://schema.org/identifier",
+                    "property": "vocab:Drone",
                     "readonly": "false",
                     "required": "true",
-                    "title": "DroneID",
+                    "title": "DroneURI",
                     "writeonly": "false"
                 }
             ],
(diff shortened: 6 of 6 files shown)


chrizandr:
So I did a little digging around in the spec to see how we can solve this problem, and I did get a possible solution. Basically the spec mentions that a collection is a subclass of hydra:Resource. This would imply that a collection will have all the fields that a typical resource has: @id, supportedProperties, supportedOperations. I've been speculating for some time now that our treatment of collections as a single resource for every class is wrong. Nowhere in the spec is it mentioned that a collection is a set of all objects for a given class. It is mentioned that a collection is "a set of somehow related resources". This does not imply that a collection would be only limited to class types. I think the direction we must move in is for users to be able to define collections on their own. Each collection itself would have an `@id` since it is a subclass of hydra:Resource. We then use the collection endpoint to relate classes/properties. 

A typical example would be:
```
{
    @id: /api/post/7
    comments: /api/commentCollection/21
}


...

{
    @id: /api/commentCollection/21
    members: [
         { @id: /api/comment/13
         },
         { @id: /api/comment/2
         },
          ...
    ]
}
```

We should have a discussion on this and decide.

vddesai1871:
@chrizandr Thanks for the explanation you gave on the call.
The main issue I am facing here is how such properties are defined in the ApiDoc, the `expects` and `returns` fields which we talked about are related to operations, not the properties. Whichever approach we use at hydrus, we will need to first settle how such relations are defined in the ApiDoc. I faced it as a major issue when I tried to implement this feature. I'll try to explain with an example.
As Akshay explained in the issue #386 a `comment`(object of a nested collection) will have a property pointing to the parent element(article on which this comment was posted), this property will be used by hydrus when new comment is posted (in our current implementation we need it to update the children list of that article), So in that case this parent_id field in the comment class is of utmost importance. When new comment object is added, hydrus needs to parse it's body and get the parent_id, but how hydrus finds out which property holds the value of the `parent_id`? it can get that detail only from the ApiDoc. Comment class's supportedProperty will have an array of properties, one of these properties is about parent_id, now the question is how it is defined so it can be used to deduce that this property holds id of the parent object?
This property might look something like
```
{
                    "@type": "SupportedProperty",
                    "property": "http://schema.org/identifier",
                    "readonly": "false",
                    "required": "true",
                    "title": "ArticleID",
                    "writeonly": "false"
},
```
But it does not provide enough details to deduce that `ArticleID` property holds the value of the parent's id. Having `"property": "vocab:Article"` won't work as we are talking about object id not the whole object.
Only possible way I can think of to solve this is what Markus uses here http://www.markus-lanthaler.com/hydra/api-demo/vocab
```
{
                    "property": {
                        "@id": "vocab:Comment/issue",
                        "@type": "hydra:Link",
                        "label": "issue",
                        "description": "The issue this comment belongs to",
                        "domain": "vocab:Comment",
                        "range": "vocab:Issue",
                        "supportedOperation": [
                            {
                                "@id": "_:issue_retrieve",
                                "@type": "hydra:Operation",
                                "method": "GET",
                                "label": "Retrieves a Issue entity",
                                "description": null,
                                "expects": null,
                                "returns": "vocab:Issue",
                                "statusCodes": []
                            },
                            {
                                "@id": "_:issue_replace",
                                "@type": "hydra:Operation",
                                "method": "PUT",
                                "label": "Replaces an existing Issue entity",
                                "description": null,
                                "expects": "vocab:Issue",
                                "returns": "vocab:Issue",
                                "statusCodes": [
                                    {
                                        "code": 404,
                                        "description": "If the Issue entity wasn't found."
                                    }
                                ]
                            }
                        ]
                    },
                    "hydra:title": "issue",
                    "hydra:description": "The issue this comment belongs to",
                    "required": null,
                    "readonly": true,
                    "writeonly": false
                },
```
You can see how `property` field of `SupportedProperty` is defined with help of `hydra:Link`.  The same  appraoch is used to define proeprty which leads to nested collection. If something like this is documented in the ApiDoc then hydrus can easily use that to handle such parent-child relations.
So can you take a look at `hydra:Link`? For whatever reason I think it is not possible to implement this functionality without incorporating `hydra:Link`.

chrizandr:
I think there is some misunderstanding about what I mentioned above. This will be a long post so please take your time to read it. There are basically two ways in which we define parent child relations. Let us take Markus's API for example. We have Issues(parent) and Comments(child). 

Now the way Markus defined his API is using a `parent_id`, which he calls `issue` for every child instance to define the parent child relation. This `parent_id` is stored as property with `@type: hydra:Link` because it is basically a URI rather than an instance of the parent. In your example you used `schema.org/identifier` instead of `hydra:Link`. This is still one way of defining this relation, i.e. using `parent_id` on every child object. There are some problems that we discussed with this approach, which is mainly related to retrieving the Comments for a particular issue(we will have to search all comments for every request.). 

Akshay gave an alternate approach to remove the problem of searching. He suggested storing child_id on every parent. To do this, he suggested that we store a list of the child_id's in every Issue object. This list can be `https://schema.org/ItemList` or something similar that we decide upon. Now this is a second kind of approch where we have child_id instead of parent_id. 

The thing that I suggested was that we use Collections directly instead of having pseudo-Collections defined by parent_id or child_id. The main problem with both the above approaches is as you mentioned, identifying the property which contains the parent/child id. This is because `schema.org/identifier` or `hydra:Link` and even `https://schema.org/ItemList` are general definitions that can be used for any property and need not necessarily be for only parent/child ids. The advantage of using a collection is that we define a CommentCollection object, where we link all Comment objects for a particular issue. We then give that CommentCollection object as the property for the Issue. In terms of defining it in the API documentation, it would be something like:
```
{
        "@type": "SupportedProperty",
        "property": vocab:CommentCollection,
        "readonly": "false",
        "required": "true",
        "title": "Comments",
        "writeonly": "false"
}
```
The property would map to an instance of the CommanCollection class. This will be defined in the supportedProperty field for the definition of the Issue class. Like any other object, when we define the Issue object, we will add the appropriate instance of the CommandCollection to the Issue instance. So an object would be something like:

```
{
        "@type": "Issue",
        "@id": "/api/Issues/27",
        "Issue": "....",
        "Comments": "/api/CommentCollection/32",
        ....
}
```

And then `/api/CommentCollection/32` would have:

```
{
        "@type": "CommentCollection",
        "@id": "/api/CommentCollection/32",
        "members": [
                {
                        "@id": "/api/Comment/12",
                        "@type": "Comment"
                },
                {
                        "@id": "/api/Comment/18",
                        "@type": "Comment"
                },
                {
                        "@id": "/api/Comment/22",
                        "@type": "Comment"
                }
        ]

}
```
Keep in mind that `/api/CommentCollection` would still contain the super set of all comments. Now when a user wants to add a new comment related to a given issue, he pushes to the CommentCollection related to that issue. We can constraint that comments cannot be created without belonging to a sub collection using readonly, writeonly properties on the Comment class. If it is readonly, then you cannot create a Comment using '/api/Comment/', you have to use `/api/CommentCollection/<id>` to create one. 

We don't use hydra:Link in this case and then we don't really have to search for properties containing hydra:Link(which can include properties other than parent/child id). To implement this however, we will have to redifine how we use Collections. We allow users to create collections(this is not there in hydrus as of now). As per the current design, PUT/POST/DELETE on the CommentCollection endpoint will create, update, delete Comment instaces. What I am suggesting is that we use this to create, update, delete CommentCollections each with a separate `@id` and `members`. I think this would be as per the spec, given that Collections are not just super sets, but related objects. There can be multiple relations defined for objects of the same type.

I hope this clears things up a bit. We can discuss more in the call later.

xadahiya:
I agree! This can be a much better approach if implemented properly.
Using the example Chris used above, if `/api/CommentCollection/` will return the superset of all the comments then what will `/api/Comment/` return? 
Should leave it blank or make in inaccessible? If yes, how will the user define all this while creating the hydra doc? Since hydrus is supposed to be started by using a HydraDoc only.

Also, it makes more sense that the members of  `/api/CommentCollection/` extend the same URI structure instead of using a custom one (`/api/Comment/1231 in this case). We should keep in mind this is a general-purpose server and defining the above-mentioned structure using HydraDoc will be quite difficult for the users. 

Keep all these things in mind, I think we should make the following adjustment to the approach proposed by @chrizandr above.
- `/api/Comment/` will be a simple comment collection endpoint.
- `/api/CommentCollection` will be a collection of collections i.e each member will be a collection consisting of a subset of comments (This is relatively easy to define in Hydra Vocab)
- `/api/CommentCollection/123` will return a comment collection specific to an issue.
- We can disable item creation at `/api/Comment` endpoint will only support `POST`, `GET` and `DELETE` operations
- When a new issue is created, a `/api/CommentCollection/<id>` will be assigned automatically to the issue and this `URI` can then be used to create to comments to that issue.
- This endpoint will only support `GET`, and `PUT` (`DELETE` can be discussed)
- /api/CommentCollection` will only support `GET`, creation and deletion of new members will be handled internally with the creation and deletion of new issues   

One more thing that is open to discussion is whether we should create `/api/Comment/` collection endpoint or not. The alternative approach can be
- ` /api/CommentCollection` - All it's members will be collections of comments
-  /api/CommentCollection/123` - It'll be a collection and all its members will be comments 
- ` /api/CommentCollection/123/<id>` - Will be a comment

The URI hierarchy in this approach will be much more simpler.
@chrizandr ^^

chrizandr:
Thanks Akshay for the comments. There are some issues I see though with this approach. Let me address them first and then maybe we can work out a solution:

> - /api/Comment/ will be a simple comment collection endpoint.
> - /api/CommentCollection will be a collection of collections i.e each member will be a collection consisting of a subset of comments (This is relatively easy to define in Hydra Vocab)
In the API Doc, the class Comment is supposed to be a Comment rather than a CommentCollection. Would it be okay to name the endpoint as Comment and return a CommentCollection instead? Same thing for the CommentCollection endpoint, which should return a collection of Comments not a collection of CommentCollection. In fact we don't define a collection of CommentCollection in the API Doc itself, that would have to be CommentCollectionCollection. This approach seems to be going against the API Doc definitions of the classes right?

> - We can disable item creation at /api/Comment endpoint will only support POST, GET and DELETE operations
At what endpoint would we then create new comments? `/api/CommentCollection/` would be for creating CommentCollections right? Do you suggest using `/api/CommentCollection/<id>` to create new comments?

> When a new issue is created, a /api/CommentCollection/<id> will be assigned automatically to the issue and this URI can then be used to create to comments to that issue.
> This endpoint will only support GET, and PUT (DELETE can be discussed)
> /api/CommentCollection will only supportGET`, creation and deletion of new members will be handled internally with the creation and deletion of new issues
I think we are getting too specific here with regards to the example. Generally we should not create any endpoints or constraints of our own. We let the user decide on the implementation that is needed based on the API. For example, if we have a property named `comments` with `required=True` for every Issue, then we don't need to create a collection automatically. Any issue without a CommentCollection would be rejected. Same thing about CommentCollection only supporting GET, these would be defined in the API Doc, so we can't really enforce this based on the Comment/Issue example.

Basically we can't really enforce such constraints since they are defined in the API Doc provided by the user. We can only decide what the output format of GET would be, since that is not in control of the user. 

A much simpler approach in my opinion would be:
- `/api/Comment/` to get/create Comment.
- `/api/Comment/<id>` to update/delete Comment
- `/api/CommentCollection/` to get/create CommentCollection. [GET will get the superset collection]
- `/api/CommentCollection/<id>` to update/delete CommentCollection.

This is similar to what hydrus uses now, with just the addition of  `/api/CommentCollection/<id>` endpoint. If we want a collection of CommentCollection, then we would have to define CommentCollectionCollection in the API Doc as well. So we leave that to the user instead. 
One thing that needs to be done though for now is that in the core repo HydraCollection will now inherit HydraClass. This would allow HydraCollection to have collections as well, which would basically allow a collection of HydraCollections, but we leave that to the developer to implement it specific to the API.

xadahiya:
> In the API Doc, the class Comment is supposed to be a Comment rather than a CommentCollection. Would it be okay to name the endpoint as Comment and return a CommentCollection instead? Same thing for the CommentCollection endpoint, which should return a collection of Comments not a collection of CommentCollection. In fact, we don't define a collection of CommentCollection in the API Doc itself, that would have to be CommentCollectionCollection. This approach seems to be going against the API Doc definitions of the classes right?

CommentCollectionCollection LOL! I don't think there will be an issue at this part since we have the functionality to name the CollectionClasses whatever we want in Doc Writer.

> At what endpoint would we then create new comments? /api/CommentCollection/ would be for creating CommentCollections right? Do you suggest using /api/CommentCollection/<id> to create new comments?

Yes!

> I think we are getting too specific here with regards to the example. Generally we should not create any endpoints or constraints of our own. We let the user decide on the implementation that is needed based on the API. For example, if we have a property named comments with required=True for every Issue, then we don't need to create a collection automatically. Any issue without a CommentCollection would be rejected. Same thing about CommentCollection only supporting GET, these would be defined in the API Doc, so we can't really enforce this based on the Comment/Issue example.
Basically we can't really enforce such constraints since they are defined in the API Doc provided by the user. We can only decide what the output format of GET would be, since that is not in control of the user.

I agree!

> A much simpler approach in my opinion would be:
    /api/Comment/ to get/create Comment.
    /api/Comment/<id> to update/delete Comment
    /api/CommentCollection/ to get/create CommentCollection. [GET will get the superset collection]
    /api/CommentCollection/<id> to update/delete CommentCollection.

The approach you've suggested is good but if we add comments at `/api/Comment` then how will hydrus know which CommentCollection to add the comment to?
Also, who will create CommentCollections and when will they be created?

>   /api/CommentCollection/ to get/create CommentCollection. [GET will get the superset collection]
The issue here is, if this endpoint shows a all comments as its members then it doesn't make much sense to go down the URI hierarchy and return something else. (CommentCollection) in this case

Thoughts? @chrizandr 
Maybe @vddesai1871 can weigh in and use this discussion to propose a concrete solution?

Mec-iS:
> He suggested storing child_id on every parent

this would be computationally too expensive.

Remember that Hydra is an hypermedia tool so this problem should be solved using a single `Hydra:Link` or a round trip of links. what is the problem with storing a `parent_id` every time a new child is stored? hydrus needs to adhere to the spec but can also establish best practices according to it.

vddesai1871:
@chrizandr @xadahiya Thanks for your help!
>hydra:Link(which can include properties other than parent/child id). 

Yeah, that's right. I didn't mean that every `hydra:Link` property will be treated as parent/child id. What I meant was that hydrus can use these Links to find those parent/child relationships indirectly. It'll be a general solution, given that ApiDoc designer uses the approach used by Markus to define such relations. And I think that approach is a really good one as it gives all control to the API designer. 

### Adding child object

Whenever a new object is added, hydrus will check if the Class of that object(`Comment` here) has any `hydra:Link` properties, if there is such property with title `xyz` then the range of that `hydra:Link` property will be something like `vocab:parentClass`(`vocab:Issue` here), then hydrus will look at `Issue` class and its supported properties, if `Issue` has a `hydra:Link` property with range as Collection of the child class whose object is being added(`CommentCollection` here.) then it's obvious that this particular property `xyz` is the `parent_id` property with link to the parent object. Then hydrus can extract the id of the parent object and increment its children list. It'll be a required property obviously as @xadahiya pointed out here https://github.com/HTTP-APIs/hydrus/issues/386#issuecomment-508390329

### Retrieving parent object
When client makes a request for the parent object(particular `Issue`),  while creating response body hydrus checks for `hydra:Link` property with childCollection as range, on finding one it will fill that field in response body as `children: "api/parent/id/childCollection"`
for example
```
{
    "@id": "api/articles/2",
    "comment": "api/articles/2/comments",
...
}
```
We can use any URI format as client will know that this field will have Link to commentCollection.
Client will dereference it to get those comments. 
As a response of `api/articles/id/commets` hydrus will return a collection of comment, the members will belong to the orginal commentCollections.
ex.
```
{
    @id: /api/articles/2/comments
    members: [
         { @id: /api/comments/13
         },
         { @id: /api/comments/2
         },
          ...
    ]
}
```
So that's basically what I was proposing based on what we disscused here https://github.com/HTTP-APIs/hydrus/issues/386

We will need to add support for `hydra:Link` for that.
in doc_writer we will need to create a class for HydraLink.
HydraSupportedProp right now only takes string as `property`, now it will take string or HydraLink. Simmilar chnages will be required in doc_maker to create HydraLink from doc.jsonld as add to HydraSupportedProp in ApiDoc.

Why API developer will choose to use something like CommentCollectionCollection over utilizing hydra:Link. :smile: 
So I would like if you guys consider it too. Otherwise we can go with what you are proposing.
I have pretty much clear understanding till here https://github.com/HTTP-APIs/hydrus/pull/406#issuecomment-513076763

/api/Comment/ to get/create Comment.
/api/Comment/<id> to update/delete Comment

vddesai1871:
>>

    He suggested storing child_id on every parent

>this would be computationally too expensive.

How so? parent contains only one field with list of children ids in string as suggested by @xadahiya(https://github.com/HTTP-APIs/hydrus/issues/386#issuecomment-508390329).
Whenever new object is added this field is updated. When serving the nestedCollection we have to just destringify that field and create appropriate URIs for those child elements.

Mec-iS:
> Whenever new object is added this field is updated. When serving the nestedCollection we have to just destringify that field and create appropriate URIs for those child elements.

This is error-prone and using dynamic data structures for potentially huge sequences (a parent can have hundred of thousands of children potentially) is too costly both now and in perspective.

vddesai1871:
>This is error-prone and using dynamic data structures for potentially huge sequences (a parent can have hundred of thousands of children potentially) is too costly both now and in perspective.


>what is the problem with storing a parent_id every time a new child is stored? 

We can go ahead with using parent_id...

xadahiya:
I agree with Lorenzo here, the approach I proposed earlier is not the right way to go. 

I think we should give it a little more thought and then decide. I believe the approach discussed above by chris and I is somewhat in the right direction. It does need some refining though.

@Mec-iS thoughts?

xadahiya:
> We can go ahead with using parent_id...

This will be super expensive since we'll have to go through the entire database.

Mec-iS:
> This will be super expensive since we'll have to go through the entire database.

Just set an index on parent_id then, that is what database indexes are for

vddesai1871:
`Instance` model in hydrus will have extra `parent_id` field, usually none, set only for child objects. If we set index on it we can directly search for all insances with parent_id = id quickly.
`instances = session.query(Instance).filter(
Instance.parent_id == id_).all()`

I don't think it would be too expensive

xadahiya:
Didn't think about indexes. LOL!
> If we set index on it we can directly search for all insances with parent_id = id quickly.

This should work! @chrizandr ^^

chrizandr:
That still doesn't solve the issue of finding the property which contains the parent ID. 

And again this would be highly constraining where APIs that are structured with child ids would have to convert all their data to be able to use hydrus if we impose the parent id model.

I still believe the most generic implementation is using Collections. Both parent and child ids are basically defining a collection. Implementing the collection approach also opens up a more wider application area where we can define relations other than parent-child ones using Collections. If we take the parent/child id approach, any other relations we define at a later stage would require some new ID or property to be added that we will have to impose on the APIs, leading to more constriction.

vddesai1871:
>That still doesn't solve the issue of finding the property which contains the parent ID.

Won't this work?
`Whenever a new object is added, hydrus will check if the Class of that object(Comment here) has any hydra:Link properties, if there is such property with title xyz then the range of that hydra:Link property will be something like vocab:parentClass(vocab:Issue here), then hydrus will look at Issue class and its supported properties, if Issue has a hydra:Link property with range as Collection of the child class whose object is being added(CommentCollection here.) then it's obvious that this particular property xyz is the parent_id property with link to the parent object.`

>And again this would be highly constraining where APIs that are structured with child ids would have to convert all their data to be able to use hydrus if we impose the parent id model.

Yep that's true

xadahiya:
@vddesai1871 @chrizandr status?

chrizandr:
Moved to #416

PR #409, opened 2019-07-11
Add mechanism to support client-controlled pagination

### LINKED ISSUE #302: Set up "hydra:search" endpoints to filter items/ids (IRITemplate)
### I'm submitting a 
- [x] feature request.

### Current Behaviour:
As discussed in this issue in the Specs https://github.com/HydraCG/Specifications/issues/168#issuecomment-404415370
hydrus should provide an endpointbased on `hydra:search` to introspect/filter items and ids. In particular:

### Expected Behaviour:
As described in the issue mentioned above:
> A better way for the client to discover the ID of an object would be, IMHO, to query the server for that object based on functional dependencies. E.g.: find the book with ISBN 123456 ; find all sensors on network X with serial number Y. That way, the client does not have to know how the server generates ID, and the server does not have to commit forever to a given way of generating IDs. All the server has to do is to provide the "search" functionality described above.

vddesai1871:
Fixes #302 

### Checklist

### Description

Refactors some old code and provides clients way to control pagination parameters `pageIndex`, `limit` and `offset`.

### DIFF
--- a/hydrus/data/crud.py
+++ b/hydrus/data/crud.py
@@ -46,7 +46,8 @@ from hydrus.data.exceptions import (
     InstanceNotFound,
     PageNotFound,
     InvalidSearchParameter,
-    IncompatibleParameters)
+    IncompatibleParameters,
+    OffsetOutOfRange)
 # from sqlalchemy.orm.session import Session
 from sqlalchemy.orm.scoping import scoped_session
 from typing import Dict, Optional, Any, List, Tuple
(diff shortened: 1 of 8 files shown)


vddesai1871:
>Now we have pageIndex, limit and offset, right?

Yes.
> Can you write a small modification in readme on how to use them properly so I can test? Was looking for that feature

Sure, once it's reviewed I'll write proper documentation. Meanwhile, recent comments posted on #302  might be helpful to you.

hydrus/data/crud.py
```diff
@@ -570,12 +564,16 @@ def get_collection(API_NAME: str,
...
-
-    # To paginate, calculate offset and page_limit values for pagination of search results
-    page_limit, offset = calculate_page_limit_and_offset(paginate=paginate, page_size=page_size, page=page,
-                                                         result_length=result_length)
-
-    for i in range(offset, offset+page_limit):
+    try:
+        # To paginate, calculate offset and page_limit values for pagination of search results
+        page, page_size, offset = pre_process_pagination_parameters(search_params=search_params, paginate=paginate,
+                                                                 page_size=page_size, result_length=result_length)
```
Guttz: This line is 116 characters long, isn't the PEP8 test supposed to fail with this?

PR #414, opened 2019-07-23
Add support for client-server synchronization

### LINKED ISSUE #300: Push notifications on data changes (client-server data sync)
### I'm submitting a 
- [x ] feature request.

### Current Behaviour:
There is no way for now to make the client-side graph representation of the data aware of data changes. Clients have to query the server every time to be sure that data has not been updated since the last query. This opens to the possibility of the client using stale data. 

### Expected Behaviour:
hydrus should allows the clients to connect through  WebSocket to push data changes for every object directly to the client.
Maybe also needed some sort of security mechanism should be present; server keeps a table for its outbox and client keeps an inbox.  The server keeps a log of changes in an outbox table, the client every time before sending a request should check the server's outbox if its inbox is synced with the latest changes. This implies that every request is under-the-hood two requests.
This refers to #218 as the inbox/outbox may be a DAG.

This will probably be a task to be part of GSOC 2019

Please start collecting here ideas for implementation, tell the community how you would like to see implemented this feature.

vddesai1871:
Fixes #300 

### Checklist

### Description

Implements socket functionality to propagate changes to all socket connected clients.
Adds `modification-table-diff` endpoint to help clients reconnect.

### Missing bits
1. Handling modifications done by multiple deletes.
- In current design `resource_url` holds the URL of the modified resource, in the case of multiple_deletes, there will be multiple such URLs. One way is to add an extra field `resource_url_set` and use it for such cases but we might need to store it as a string and as discussed earler that may not be the best way to handle it. So any other approach you can suggest...?
2. @Guttz From what I get, the client always uses its `agent_job_id` while reconnecting, so it can pick up from where it left, but when the `agent_job_id` provided by the agent is too old, hydrus won't be able to find it in its modification table(as it has limit of 1000 entries or some another limit), so in that case hydrus (temporarily) returns an empty list(as you are accepting something similar to empty list at client by checking for `not new_rows`). Instead of an empty list hydrus can use some status object or something else, I am not sure about which status_code to use here but something else than an empty list would be better to show that client has been outdated.

### DIFF
--- a/hydrus/app_factory.py
+++ b/hydrus/app_factory.py
@@ -3,7 +3,7 @@ from flask_cors import CORS
 from flask_restful import Api
 
 from hydrus.resources import (Index, Vocab, Contexts, Entrypoint,
-                              ItemCollection, Item, Items)
+                              ItemCollection, Item, Items, ModificationTableDiff)
 
 
 def app_factory(api_name: str = "api") -> Flask:
@@ -38,5 +38,9 @@ def app_factory(api_name: str = "api") -> Flask:
         "/{}/<string:path>/add/<int_list>".format(api_name),
         "/{}/<string:path>/add".format(api_name),
         "/{}/<string:path>/delete/<int_list>".format(api_name))
+    api.add_resource(
+        ModificationTableDiff,
+        "/{}/modification-table-diff".format(api_name),
+        endpoint="modification_table_diff")
 
     return app
--- a/hydrus/app_factory.py
+++ b/hydrus/app_factory.py
@@ -3,7 +3,7 @@ from flask_cors import CORS
 from flask_restful import Api
 
 from hydrus.resources import (Index, Vocab, Contexts, Entrypoint,
-                              ItemCollection, Item, Items, ModificationTableDiff)
+                              ItemCollection, Item, Items)
 
 
 def app_factory(api_name: str = "api") -> Flask:
@@ -38,9 +38,5 @@ def app_factory(api_name: str = "api") -> Flask:
         "/{}/<string:path>/add/<int_list>".format(api_name),
         "/{}/<string:path>/add".format(api_name),
         "/{}/<string:path>/delete/<int_list>".format(api_name))
-    api.add_resource(
-        ModificationTableDiff,
-        "/{}/modification-table-diff".format(api_name),
-        endpoint="modification_table_diff")
 
     return app
--- a/hydrus/data/crud.py
+++ b/hydrus/data/crud.py
@@ -744,7 +744,9 @@ def get_modification_table_diff(session: scoped_session,
                 Modification.job_id == agent_job_id).one()
         except NoResultFound:
             return []
+    # Otherwise
     try:
+        # Get all modifications in sorted order from latest to oldest
         modifications = session.query(Modification).order_by(Modification.timestamp.desc()).all()
     except NoResultFound:
         modifications = []
--- a/hydrus/data/crud.py
+++ b/hydrus/data/crud.py
@@ -36,7 +36,7 @@ from sqlalchemy.orm import with_polymorphic
 from sqlalchemy import exists
 from sqlalchemy.orm.exc import NoResultFound
 from hydrus.data.db_models import (Graph, BaseProperty, RDFClass, Instance,
-                                   Terminal, GraphIAC, GraphIIT, GraphIII)
+                                   Terminal, GraphIAC, GraphIIT, GraphIII, Modification)
 from hydrus.data.exceptions import (
     ClassNotFound,
     InstanceExists,
@@ -737,6 +737,39 @@ def delete_single(type_: str, session: scoped_session) -> None:
         raise InstanceNotFound(type_=rdf_class.name)
 
     return delete(instance.id, type_, session=session)
+
+
+def insert_modification_record(method: str, resource_url: str,
+                               session: scoped_session) -> str:
+    """
+    Insert a modification record into the database.
+    :param method: HTTP method type of related operation.
+    :param resource_url: URL of resource modified.
+    :param session: sqlalchemy session.
+    :return: ID of new modification record.
+    """
+    modification = Modification(method=method, resource_url=resource_url)
+    session.add(modification)
+    record_count = session.query(Modification).count()
+    if record_count > 1000:
+        oldest_record = session.query(Modification).order_by(Modification.timestamp.asc()).first()
+        session.delete(oldest_record)
+    session.commit()
+    return modification.job_id
+
+
+def get_last_modification_job_id(session: scoped_session) -> str:
+    """
+    Get job id of most recent modification record stored in the db.
+    :param session: sqlalchemy session
+    :return: job id of recent modification.
+    """
+    last_modification = session.query(Modification).order_by(Modification.timestamp.desc()).first()
+    if last_modification is None:
+        last_job_id = ""
+    else:
+        last_job_id = last_modification.job_id
+    return last_job_id
 
 
 def apply_filter(object_id: str, session: scoped_session,
--- a/hydrus/data/crud.py
+++ b/hydrus/data/crud.py
@@ -772,6 +772,39 @@ def get_last_modification_job_id(session: scoped_session) -> str:
     return last_job_id
 
 
+def get_modification_table_diff(session: scoped_session,
+                                agent_job_id: str = None) -> List[Dict[str, Any]]:
+    """
+    Get modification table difference.
+    :param session: sqlalchemy session.
+    :param agent_job_id: Job id from the client.
+    :return: List of all modifications done after job with job_id = agent_job_id.
+    """
+    # When client provides a job id but server does not have corresponding record
+    # meaning that client is too outdated.
+    if agent_job_id is not None:
+        try:
+            record_for_agent_job_id = session.query(Modification).filter(
+                Modification.job_id == agent_job_id).one()
+        except NoResultFound:
+            return []
+    try:
+        modifications = session.query(Modification).order_by(Modification.timestamp.desc()).all()
+    except NoResultFound:
+        modifications = []
+    list_of_modification_records = []
+    for modification in modifications:
+        if modification.job_id == agent_job_id:
+            break
+        modification_record = {
+            "job_id": modification.job_id,
+            "method": modification.method,
+            "resource_url": modification.resource_url
+        }
+        list_of_modification_records.append(modification_record)
+    return list_of_modification_records
+
+
 def apply_filter(object_id: str, session: scoped_session,
                  search_props: Dict[str, Any]) -> bool:
     """Check whether objects has properties with query values or not.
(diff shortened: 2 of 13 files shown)


Guttz:
Regarding the second issue, I agree that a status code is a better solution. Regarding the code I would suggest 204:

**204 No Content - The server successfully processed the request and is not returning any content.**

About the first one, I didn't quite get the problem. Could you detail more?

Guttz:
I tested both the Agent and hydrus together and it's working quite well!! :+1: 

Agent automatically connects after fetching the API Doc and listen to all updates and deletions and instantly updates its internal representation as we wanted. When implementing the data inside the socket as you suggested there's only one small modification I had to make I'll point out below.

hydrus/helpers.py
```diff
@@ -239,3 +239,21 @@ def add_pagination_iri_mappings(template: str,
...
+                     method: str, resource_url: str):
+    """Sends synchronization update to all connected clients.
+    :param socketio: socketio connection.
+    :param new_job_id: Job id of the new modification(update).
+    :param last_job_id: Job id of the last(most recent) modification until this new one.
+    :param method: Method type of the operation.
+    :param resource_url: URL of resource which needs to be synchronized.
+    """
+    data = {
+        "new_job_id": new_job_id,
```
Guttz: Can you change` new_job_id `to `job_id` here in the dict? Just so the dict for a new row is the same you return in when querying by the /modification-table-diff endpoint. 
`"job_id": new_job_id,`

This was the only small change I had to make to make everything work.

hydrus/data/crud.py
```diff
@@ -695,6 +695,74 @@ def delete_single(type_: str, session: scoped_session) -> None:
...
+    :param session: sqlalchemy session.
+    :param agent_job_id: Job id from the client.
+    :return: List of all modifications done after job with job_id = agent_job_id.
+    """
+    # When client provides a job id but server does not have corresponding record
+    # meaning that client is too outdated.
+    if agent_job_id is not None:
+        try:
+            record_for_agent_job_id = session.query(Modification).filter(
+                Modification.job_id == agent_job_id).one()
```
xadahiya: This never gets returned. What's the use?

xadahiya: Shouldn't this try to return every element starting from the job_id agent sent? @Guttz @vddesai1871

hydrus/helpers.py
```diff
@@ -239,3 +239,21 @@ def add_pagination_iri_mappings(template: str,
...
+        "last_job_id": last_job_id,
```
xadahiya: Why are we sending the last_job_id?

hydrus/resources.py
```diff
@@ -154,8 +156,16 @@ def post(self, id_: str, path: str) -> Response:
...
+                            get_hydrus_server_url(), get_api_name(), path, object_id)
+                    last_job_id = crud.get_last_modification_job_id(session=get_session())
+                    new_job_id = crud.insert_modification_record(method, resource_url,
+                                                                 session=get_session())
+                    send_sync_update(socketio=socketio, new_job_id=new_job_id,
+                                     last_job_id=last_job_id, method=method,
+                                     resource_url=resource_url)
+                    headers_ = [{"Location": resource_url}]
                     status_description = "Object with ID {} successfully updated".format(object_id)
                     status = HydraStatus(code=200, title="Object updated", desc=status_description)
```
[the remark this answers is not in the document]
vddesai1871: I don't know whether client caches collections or not, nevertheless I made a commit adding sync update for PUT operation over collections.

hydrus/data/crud.py
```diff
@@ -695,6 +695,74 @@ def delete_single(type_: str, session: scoped_session) -> None:
...
+    return modification.job_id
+
+
+def get_last_modification_job_id(session: scoped_session) -> str:
+    """
+    Get job id of most recent modification record stored in the db.
+    :param session: sqlalchemy session
+    :return: job id of recent modification.
+    """
+    last_modification = session.query(Modification).order_by(Modification.timestamp.desc()).first()
```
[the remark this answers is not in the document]
vddesai1871: I haven't considered it yet, it would greatly simplify the process to get last_job_id by directly using `new_job_id-1`.

hydrus/data/crud.py
vddesai1871: It was for checking whether the client is too outdated or not(if the client tries to get all modification after some `agent_job_id`=`x` and if `x` is not in hydrus DB then it means the client is too outdated).
But yeah it was too complicated, I have refactored it to increase readability.

hydrus/helpers.py
vddesai1871: https://github.com/HTTP-APIs/hydra-python-agent/pull/123#discussion_r305711803

hydrus/resources.py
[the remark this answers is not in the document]
Guttz: I thought a lot about this too and think it's an important architectural discussion I wanted to have.

At first, I also wanted to add PUT to sync however we're using the Agent to cache resources, in that sense, syncing PUT requests would be using Redis more in a "db mirror/copy way" more than caching. I think this architectural decision of syncing PUT requests would be really important if we didn't have IRI Template to search the server.

I think both approaches have different benefits and it depends on the user use case. When imagining a use case where the Server has millions of resources and the Agent is a regular application that, for example, fetches the server with a 5-sized-pagination-size, always copying the whole database then syncing PUT will generate a huge amount of requests which won't be useful and ever used by the user. However, only caching the resources he actually queries for will produce 1 request for each and them save a lot of future requests for that resource since it's cached.

>  "check if there is an item in the collection with name xyz" then what will the agent do

When fetching a hydrus collection it returns a pagination-sized list of "@id" and "@type", so it doesn't have the members details to already cache these resources. So If a user queries "check if there is an item in the collection with name xyz", by default, what will happen is that the Agent will forward the IRI Template that will search the DB internally and return the collection with a list having "@id" and "@type" for the found resources. This list can be used to fetch the resources itself using their ID, which might be already cached or will be cached if not. In that scope, **I still wanted to have a way to cache Search Queries** and not go for the server always.. And for that, I came up with the cached_limit optional parameter so the user can inform how many items he wants to fetch and, if internally Redis has enough resources for that search, it won't go for the server, it will return from the already cached resources.

Give me your feedback on that

hydrus/data/crud.py
```diff
@@ -695,6 +695,74 @@ def delete_single(type_: str, session: scoped_session) -> None:
...
+    :param method: HTTP method type of related operation.
+    :param resource_url: URL of resource modified.
+    :param session: sqlalchemy session.
+    :return: ID of new modification record.
+    """
+    modification = Modification(method=method, resource_url=resource_url)
+    session.add(modification)
+    record_count = session.query(Modification).count()
+    if record_count > 1000:
+        oldest_record = session.query(Modification).order_by(Modification.timestamp.asc()).first()
```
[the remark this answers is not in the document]
vddesai1871: *Work in progress*

hydrus/resources.py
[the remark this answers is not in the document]
Mec-iS: we are not really interested il storing collections in redis as we should have already all their elements as nodes in the graph. so i would focus in keeping instances updated and just assume that the collection is the output of querying all the nodes of a given type

hydrus/helpers.py
xadahiya: Okay cool! 👍

hydrus/data/crud.py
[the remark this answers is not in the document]
chrizandr: Even while using incremental ID's it's a bad idea to use `new_job_id - 1` to get the last ID, there may be deletions in the table and it is generally considered as a bad practice to do this.

hydrus/data/crud.py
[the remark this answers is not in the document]
vddesai1871: While implementing multi-threading I was facing issues with in :memory: db of sqlite3. On some cursory look-up, I found out this https://docs.sqlalchemy.org/en/13/dialects/sqlite.html#using-a-memory-database-in-multiple-threads and made it work but still, there were unexpected bugs and use of `scopped_session` was not possible. So, instead of using in :memory: DB I used the database.db file as the default DB. The current approach with `socpped_session` is the recommended approach to add multi-threading, it works for all DBs except in :memory: Sqlite.

hydrus/app_factory.py
```diff
@@ -38,5 +38,9 @@ def app_factory(api_name: str = "api") -> Flask:
         "/{}/<string:path>/add/<int_list>".format(api_name),
         "/{}/<string:path>/add".format(api_name),
         "/{}/<string:path>/delete/<int_list>".format(api_name))
+    api.add_resource(
```
chrizandr: This is not part of the API. I'm not sure if we should serve this here. Shouldn't the diff and all sync related mechanism be served using socket events? Why are we serving this information at the API endpoint. Ideally the app should contain the API and socket events should be done by using `@socket.on` decorators.

hydrus/data/exceptions.py
```diff
@@ -174,3 +174,11 @@ def get_HTTP(self) -> HydraError:
         """Return the HTTP response for the Exception."""
         description = "Offset {} is out of range.".format(self.offset)
         return HydraError(code=400, title="Page not found", desc=description)
+
+
+class ClientTooOutdated(Exception):
```
chrizandr: Again same here. Sync mechanism is not a part of Hydra, we don't use HydraError for this. This should be served using the socket, not the app.

hydrus/resources.py
```diff
@@ -551,3 +601,22 @@ def get(self, category: str) -> Response:
         else:
             error = HydraError(code=404, title="NOT FOUND", desc="Context not found")
             return set_response_headers(jsonify(error.generate()), status_code=error.code)
+
+
+class ModificationTableDiff(Resource):
```
chrizandr: This functionality should be implemented using `socketio.on` rather than app.

hydrus/tests/test_app.py
```diff
@@ -412,6 +415,87 @@ def test_endpointClass_GET(self):
                         assert "@id" in response_get_data
                         assert "@type" in response_get_data
 
+    def test_socketio_POST_updates(self):
```
chrizandr: The test should also use a socket to connect to the sync mechanism, rather than using an HTTP client.

vddesai1871:
Had to replace UUIDs by incremental IDs: Initially, I was using UUIDs and timestamps to keep track of modifications, but while testing I found out timestamps were not precise enough for our use case and they vary from DB to DB. While testing I had to make delay of 1 sec between two modifications so those modifications can get different timestamps and I can run tests(see commit https://github.com/HTTP-APIs/hydrus/pull/414/commits/bbcc0e5704138fe1203af3cb7ff6da0316015d63, in `test_modification_table_diff` on commenting out `sleep(1)` test will fail) so finally I replaced timestamp and UUIDs with incremental IDs so we do not have to presume such delays between operations.

xadahiya:
merging #414

PR #415, opened 2019-08-06
Use HydraLink to improve handling of nested class properties

vddesai1871:
Fixes https://github.com/HTTP-APIs/hydra-python-core/issues/30
Extends and improves the handling of nested class properties with the help of HydraLink.
Work in progress
Issue: Adding Link properties as GraphIII triples instead of GraphIIT triples.
Currently, it uses regex to identify links, extracts id of the nested resource and adds a GraphIII triple. I am looking for a more robust solution, I'll try out passing a list of Link properties to `crud.insert` and add a function to add GraphIII triples for GraphIII relationships defined with the help of links, so we can avoid regex and be sure about proper handling of such GraphIII relations expressed as link.
Also, need to adapt tests.

### Checklist

### DIFF
--- a/hydrus/data/crud.py
+++ b/hydrus/data/crud.py
@@ -116,8 +116,6 @@ def get(id_: str, type_: str, api_name: str, session: scoped_session,
         instance = session.query(Instance).filter(
             Instance.id == data.object_).one()
         object_template[prop_name] = instance.id
-        print("yes")
-        print(object_template[prop_name])
 
     for data in data_IIT:
         prop_name = session.query(properties).filter(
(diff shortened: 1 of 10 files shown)


vddesai1871:
I am preprocessing Link properties for basic check and passing a list of such properties to insert method so it can know which properties are `Link` properties and insert them properly as GraphIII triples(otherwise it will think the link is a normal string and save it as a Terminal and use GraphIIT triple, which would be wrong).
Also adapted tests(testing both `hydra:Link`(uses link) and "vocab:"(object directly nested) nested properties).

hydrus/data/crud.py
```diff
@@ -116,6 +116,8 @@ def get(id_: str, type_: str, api_name: str, session: scoped_session,
         instance = session.query(Instance).filter(
             Instance.id == data.object_).one()
         object_template[prop_name] = instance.id
+        print("yes")
+        print(object_template[prop_name])
```
xadahiya: Please remove redundant print statements

hydrus/data/crud.py
```diff
@@ -186,7 +189,39 @@ def insert(object_: Dict[str, Any], session: scoped_session,
                 # Adds new Property
                 session.close()
                 raise PropertyNotFound(type_=prop_name)
-
+            # For insertion in III through link
+            if prop_name in link_props:
+                regex = r'[a-z0-9]{8}-([a-z0-9]{4}-){3}[a-z0-9]{12}'
```
xadahiya: Add proper comments what are you trying to do with regex and why

hydrus/resources.py
```diff
@@ -144,15 +145,17 @@ def post(self, id_: str, path: str) -> Response:
         if checkClassOp(class_type, "POST") and check_writeable_props(class_type, object_):
             # Check if class_type supports POST operation
             obj_type = getType(class_type, "POST")
+            link_props = get_link_props(class_type, object_)
             # Load new object and type
             if validObject(object_) and object_["@type"] == obj_type and check_required_props(
-                    class_type, object_):
+                    class_type, object_) and isinstance(link_props, Dict):
```
xadahiya: Since get_link_props always return a dict, is `isinstance(link_props, Dict)` check necessary?

hydrus/data/crud.py
```diff
@@ -186,7 +189,39 @@ def insert(object_: Dict[str, Any], session: scoped_session,
...
-
+            # For insertion in III through link
+            if prop_name in link_props:
+                regex = r'[a-z0-9]{8}-([a-z0-9]{4}-){3}[a-z0-9]{12}'
+                matchObj = re.match(regex, link_props[prop_name])
+                # Link is to an instance of a collection class
+                if matchObj:
+                    try:
+                        nested_instance = session.query(Instance).filter(
+                            Instance.id == link_props[prop_name]).one()
```
xadahiya: Can you please explain what are you trying to do here?

hydrus/resources.py
vddesai1871: It returns False when type doesn't match. 
Class_type of resource provided as link should match to class_type defined as range of the link_property in the ApiDoc.

hydrus/data/crud.py
vddesai1871: Preprocessing of link properties change property value to instance_id in case of link to collection class and class_type in case of link to non-collection class, I am adding comments.

hydrus/data/crud.py
vddesai1871: Finds the instance which is represented by the client provided link(`link_props[prop_name]` contains the id of the instance after preprocessing.) If there is no such valid instance then we can raise an appropriate error and if there is such an instance then we can go ahead and add a new triple to store this relation.

hydrus/resources.py
vddesai1871: Removed that and made an explicit check for type-checking of links.

requirements.txt
```diff
@@ -26,5 +26,5 @@ six==1.11.0
 SQLAlchemy
 thespian==3.9.2
 Werkzeug
-git+https://github.com/vddesai1871/hydra-python-core@hydra_link#egg=hydra_python_core
+git+https://github.com/HTTP-APIs/hydra-python-core@develop#egg=hydra_python_core
```
Mec-iS: why we are not just merging hydra-python-core to master? @xadahiya @chrizandr

xadahiya: Done

xadahiya: @vddesai1871 please update this then we can merge

vddesai1871: Done

xadahiya:
Merging #415

PR #420, opened 2019-08-23
Improve tests

### LINKED ISSUE #317: function to generate dummy objects fails for classes with explicitly defined paths.

vddesai1871:
Fixes #317 

### Checklist

### Description

Extends the nested class test in test_crud.py
Refactors gen_dummy_object functions as discussed in #317 
Also makes changes in helper functions to use path instead of assuming path will be equal to class_type.

### DIFF
--- a/hydrus/helpers.py
+++ b/hydrus/helpers.py
@@ -100,39 +100,43 @@ def getType(class_path: str, method: str) -> Any:
     # 'vocab:' not everything will be returned.
 
 
-def checkClassOp(class_type: str, method: str) -> bool:
-    """Check if the Class supports the operation."""
+def checkClassOp(path: str, method: str) -> bool:
+    """Check if the Class supports the operation.
+    :param path: Path of the collection or non-collection class.
+    :param method: Method name.
+    :return: True if the method is defined, false otherwise.
+    """
     for supportedOp in get_doc(
-    ).parsed_classes[class_type]["class"].supportedOperation:
+    ).parsed_classes[path]["class"].supportedOperation:
         if supportedOp.method == method:
             return True
     return False
 
 
-def check_required_props(class_type: str, obj: Dict[str, Any]) -> bool:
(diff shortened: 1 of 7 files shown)


xadahiya:
@vddesai1871 Please recreate this once I create develop.

PR #431, opened 2019-12-19
Fixed authorization bug, by creating Cookie based authentication nonce

### LINKED ISSUE #430: Authentication not working.
### I'm submitting a 
- [x] bug report.
- [ ] feature request.

### Current Behaviour:

Running  `docker-compose up --build` creates the hydrus server, but when trying to login to access properties like `api/Area` with the default credentials, the alert box, either refreshes or the user has to click cancel.

### Expected Behaviour:

On entering the correct credential, user should be served with the default `Response 200` code.
else `failed to authenticate` message.

### Steps to reproduce:

Run server using
`docker-compose up --build`

then visit 
`http://localhost:8000/api/Area`

any property which requires authentication can be used to reproduce the error, not just "Area".

YES.
I have retraced the error, the problem lies in `Check_nonce` method as it is unable to find `X-Authentication` header. It can't find the `X-Authentication` header as its the part of the response headers and is given after the page has been requested. We are trying to find those headers in "pending" state of the web page.

vedangj044:
Fixes #430 

### Checklist

### Description

The current branch uses 'X-Authentication' header to create and validate the 'nonce'.
This doesn't work as `request.headers` return Request Headers of a web page, whereas 'X-Authentication' is a part of the Response Header. 

Instead of this, storing the nonce in a cookie and giving it to client, and validating the same, at the time of authentication, works fine. Using nonce as a cookie is useful to prevent replay attacks, as the nonce is valid for only a short period of time, and once validated is converted to a token.

A new nonce cookie is generated at the time of 'failed authentication'.

### Change logs

`request.set_cookie("nonce", create_nonce(session)` 
is contained by a new function named `vague_response` in `hydrus.data.user`

This function is called each time `resource.py` requests an endpoint and a cookie is not already present.

#### Changed

- `resource.py`
- `hydrus.data.user`

and some minor required changes.

 #### Removed 
`request.headers["X-Authentixation"]` and its uses

### DIFF
--- a/hydrus/app.py
+++ b/hydrus/app.py
@@ -6,7 +6,7 @@ from sqlalchemy import create_engine
 
 from gevent.pywsgi import WSGIServer
 from sqlalchemy.orm import sessionmaker
-
+from hydrus.data.db_models import User, Token, Nonce
 from hydrus.app_factory import app_factory
 from hydrus.conf import (
     HYDRUS_SERVER_URL, API_NAME, DB_URL, APIDOC_OBJ, PORT, DEBUG)
@@ -47,12 +47,13 @@ TOKEN = True
 
 if AUTH:
     try:
-        add_user(id_=1, paraphrase="test", session=session)
+        add_user(id_=5, paraphrase="test", session=session)
     except UserExists:
         pass
 
 # Create a Hydrus app
 app = app_factory(API_NAME)
+
 
 with set_authentication(app, AUTH):
     # Use authentication for all requests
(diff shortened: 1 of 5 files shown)


xadahiya:
Please create the PR in develop branch. Also, the build is failing. Please check the logs

Mec-iS:
@vedangj044 thanks for your interest and the PR.
Some hints:
* all PR need to be against `develop` branch
* when you see a red X in your commit, it means that the automatic testing (Continous Integration) for the repo has failed somehow becuase of the changes contained in the commit. To know more about the test failures you have to click on the X and the "Details" -> in this case [this link](https://github.com/HTTP-APIs/hydrus/runs/355871893). As you can read in the log, there are some tests failing and some PEP-8 warnings.

You have to be sure that the code you contributed pass all tests and warnings in the CI pipeline.

PR #432, opened 2019-12-21
Fixes HTTP-APIs#430

### LINKED ISSUE #430: Authentication not working.

vedangj044:
Fixes #430 
Corrected PEP8 issues, and 

> **Commented cookie generating function for POST and DELETE operations in a put request in ItemCollection**

### Checklist

### Description

The current branch uses 'X-Authentication' header to create and validate the 'nonce'.
This doesn't work as `request.headers` return Request Headers of a web page, whereas 'X-Authentication' is a part of the Response Header. 

Instead of this, storing the nonce in a cookie and giving it to client, and validating the same, at the time of authentication, works fine. Using nonce as a cookie is useful to prevent replay attacks, as the nonce is valid for only a short period of time, and once validated is converted to a token.

A new nonce cookie is generated at the time of 'failed authentication'.

### Change logs

`request.set_cookie("nonce", create_nonce(session)` 
is contained by a new function named `vague_response` in `hydrus.data.user`

This function is called each time `resource.py` requests an endpoint and a cookie is not already present.

#### Changed

- `resource.py`
- `hydrus.data.user`

and some minor required changes.

 #### Removed 
`request.headers["X-Authentixation"]` and its uses


xadahiya:
Authorization is working fine. Please follow this procedure http://www.hydraecosystem.org/Auth

PR #445, opened 2020-02-13
item_collection refactor

### LINKED ISSUE #424: General refactoring
### Current Behaviour:
A lot of functions have many if-else branches and the procedures implemented by these branches have too much code in them. This makes readability and testability hard.

For example the branch in this [chunk of code](https://github.com/HTTP-APIs/hydrus/blob/42205ab4d4984f591310b19f3b5d6c1610e95e4d/hydrus/resources.py#L108) should be ported into a function like:
```
    if checkClassOp(class_path, "GET"):
        items_get_check_support() 
    abort(405)
```
Into a dedicated module called `itemshelpers.py` the above functions should be defined by incorporating the current code:
```
def items_get_check_support(...):
    """ Check if class_type supports GET operation
            try:
                # Try getting the Item based on ID and Class type
                response = crud.get(
                    id_,
                    class_type,
                    api_name=get_api_name(),
                    session=get_session())

                response = finalize_response(class_path, response)
                return set_response_headers(
                    jsonify(hydrafy(response, path=path)))

            except (ClassNotFound, InstanceNotFound) as e:
                error = e.get_HTTP()
                return set_response_headers(jsonify(error.generate()), status_code=error.code)
```
Each of these helpers functions should be tested separately.

### Expected Behaviour:
All functions should perform well-scoped units of work and have their own tests.
Handlers in endpoints classes should be much more concise, all the units of work for each branch should be moved into helpers functions. 

Main functions should look something like this:
```
def post(...):
    """ Docstring """
    ... some code ...
    if condition_1:
        run_helper_1(...)
    elif condition_2:
        run_helper_2(...)
    else:
        run_helper_3(...)
    ... rest of the code ...
```

thrishik7:
Fixes #424 

### Checklist
- [ ] My branch is up-to-date with upstream/develop branch.

### Description
refactor class itemcollection

### DIFF
--- a/hydrus/itemhelpers.py
+++ b/hydrus/itemhelpers.py
@@ -197,7 +197,7 @@ def itemsCollection_get_support(collection, class_path, path, search_params):
                 return set_response_headers(jsonify(error.generate()), status_code=error.code)
 
 
-def itemsCollection_post_support(object_, link_props, path):                       
+def itemsCollection_post_support(object_, link_props, path):
     try:
         crud.update_single(
             object_=object_,
--- a/hydrus/itemhelpers.py
+++ b/hydrus/itemhelpers.py
@@ -189,8 +189,7 @@ def itemsCollection_get_support(collection, class_path, path, search_params):
                 paginate=False, path=path, search_params=search_params)
 
         response["search"] = add_iri_template(path=class_path,
-                                                API_NAME=get_api_name())
-
+                                              API_NAME=get_api_name())
         return set_response_headers(jsonify(hydrafy(response, path=path)))
 
     except (ClassNotFound, PageNotFound, InvalidSearchParameter, OffsetOutOfRange) as e:
@@ -202,8 +201,7 @@ def itemsCollection_post_support(object_, link_props, path):
     """
     Check if object_ supports POST operation
     :param object_ : endpoint or member of a collection
-    :param link_props : Dict with property_title as key and
-                        instance_id(for collection class) or class_name(for non-collection class) as value
+    :param link_props : Dict with property_title as key and instance_id(for collection class) as value
     :param path : Path for Item type ( Specified in APIDoc @id)
     """
     try:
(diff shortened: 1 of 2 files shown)


hydrus/itemhelpers.py
```diff
@@ -155,3 +170,98 @@ def items_delete_check_support(id_, class_type, path):
         error = e.get_HTTP()
         return set_response_headers(jsonify(error.generate()),
                                     status_code=error.code)
+
+
+def itemsCollection_get_support(collection, class_path, path, search_params):
+            try:
```
xadahiya: Inconsistent Indentation

hydrus/itemhelpers.py
```diff
@@ -155,3 +170,98 @@ def items_delete_check_support(id_, class_type, path):
...
+
+                return set_response_headers(jsonify(hydrafy(response, path=path)))
+
+            except (ClassNotFound, PageNotFound, InvalidSearchParameter, OffsetOutOfRange) as e:
+                error = e.get_HTTP()
+                return set_response_headers(jsonify(error.generate()), status_code=error.code)
+
+
+def itemsCollection_post_support(object_, link_props, path):
+    try:
```
xadahiya: Please add doctrings to all your functions.

xadahiya: Doctrings should be in a format similar to https://github.com/HTTP-APIs/hydrus/blob/37af5f3be07b5b21e133c40e3586f58859f38094/hydrus/helpers.py

hydrus/resources.py
```diff
@@ -64,7 +64,11 @@
     items_get_check_support,
     items_post_check_support,
```
xadahiya: Please remove all the redundant imports that are not being used here.

thrishik7:
> Great effort! However, you need to make a couple of changes before we can proceed.

okay @xadahiya , I will work on it

thrishik7:
@xadahiya  I have made the changes. Please let me know if I have to add few more changes

xadahiya:
All docstings still don't have consistent format.

thrishik7:
> All docstings still don't have consistent format.

@xadahiya okay, I will work on it.

thrishik7:
@xadahiya can you please point out what to change for anyone docstring. So that I would apply the same for the rest

xadahiya:
See `itemsCollection_get_support` for example docstring

thrishik7:
> See `itemsCollection_get_support` for example docstring
@xadahiya  should I refer `itemsCollection_get_support` and change other docstrings??

xadahiya:
All docstrings should have a consistent format like this

Issue #451, opened 2020-02-21
bug:hydrus cli takes any argument to start the server.


geekypandey:
### I'm submitting a 
- [x] bug report.
- [ ] feature request.

### Current Behaviour:

The hydrus cli function is not properly implemented. 
hydrus takes any argument and starts serving.
e.g,
`hydrus serve`
`hydrus asdfasd `
both would do the same thing.

### Expected Behaviour:

`hydrus serve` should start the server and any sub command should not be considered.

### Steps to reproduce:

after setting up the environment,try
`hydrus serve` and
`hydrus asdfasdf`
both produces same effect.

### Do you want to work on this issue?

yes!! I have already found the solution. Can I submit the PR for the same?

xadahiya:
Sure! Please go ahead

geekypandey:
I had doubts in the implementation :
1. There is open parameter defined in document as `openapi` which is not being used anywhere and not even passed as argument. Sorry but I am not yet through whole hydrus implementation , just starting to understand so don't know if that is required.
2.The implementation idea is use **serve** as an subcommand of **hydrus cli** eg .,`hydrus serve`.
And the options will be passed to the this serve subcommand and not the main hydrus cli.
3.Could I bump up the version of Click to Click==7.0 which was previously at Click==6.7 to leverage new features? I will run the tests to make sure nothing is broken because of it.

Currently I have completed the implementation as specified above.

sameshl:
@xadahiya Looks like this issue can be closed, seeing from PR https://github.com/HTTP-APIs/hydrus/pull/453

PR #453, opened 2020-02-24
FIX: issue with cli.py

geekypandey:
Fixes #

### Checklist

### Description

fixes issue #451 .
1. Fixed issue with the hydrus command line utility (cli.py)
### Change logs

1. Changed Click version from click==6.7 to click==7.0 .

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -23,12 +23,14 @@ import yaml
 from hydrus.conf import (
     APIDOC_OBJ, FOUND_DOC)
 
+
 @click.group()
 def startserver():
     """
     Python Hydrus CLI.
     """
     pass
+
 
 @startserver.command()
 @click.option("--adduser", "-u", default=tuple([1, "test"]),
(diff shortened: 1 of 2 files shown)


geekypandey:
@xadahiya @Mec-iS Could you please review?

cli.py
```diff
@@ -49,27 +57,28 @@
...
-def startserver(adduser: Tuple, api: str, auth: bool, dburl: str, pagination: bool,
-                hydradoc: str, port: int, pagesize: int, serverurl: str, token: bool,
-                stale_records_removal_interval: int, serve: None) -> None:
+def serve(adduser: tuple, api: str, auth: bool, dburl: str, pagination: bool,
+          hydradoc: str, port: int, pagesize: int, serverurl: str, token: bool,
+          stale_records_removal_interval: int) -> None:
     """
-    Python Hydrus CLI
+    Starts up the server.
+    \f
```
Mec-iS: please remove this line

Mec-iS:
Are there any tests we can add for the CLI?

cli.py
geekypandey: The line 65 would help to keep the help mode more clean for the user. We are already providing with the types of input parameter in help mode so that would be unnecessary replication.What do you think?

geekypandey:
Testing could be done for:
1. Right and wrong kind of arguments.
2. working of different arguments like `--no-auth` , `--serverurl` , `--port`.

chrizandr:
> Are there any tests we can add for the CLI?

We already have tests for CLI. These changes don't need new tests.

Issue #404, opened 2019-07-01
Creation of dynamic endpoints


vddesai1871:
### I'm submitting a 
- [ ] bug report.
- [x] feature request.

### Current Behaviour:

Right now, hydrus only serves static data and provides standard CRUD operations over
that data, it doesn’t allow the user to define some logic to process incoming data and
provide generated output.
### Expected Behaviour:

hydrus should provide some way to define dynamic endpoints, and thus let users define
logic to process incoming data.

The Command Dispatcher Pattern and can be used. The developer can create a Command_dispatcher class having commands(extra logic) as
methods ( following some protocol), then it can be passed to hydrus where hydrus will
form commands according to the protocol and check whether the given (command_dispatcher) object has that attribute or not, if yes then get that attribute (method) and execute it.
### Do you want to work on this issue?

Yes

Mec-iS:
Are you talking about a Falcon/Flask-like pre_request and post_response hooks?

vddesai1871:
>Are you talking about a Falcon/Flask-like pre_request and post_response hooks?

Yes, something very similar to that.

>This can be done programmatically using the Flask interface.

I don't know exactly how. But if we are talking about `before_request` method provided by Flask, then it applies to all requests. Which we do not want. Even if we find inbuilt support for having pre-request hook for individuals responders(like `Item/get`), it wouldn't work as `Item/get` is responsible for all hydra resources(`Drone`, `State`, etc.), and the developer might want to have pre-request functioning only for `Drone` class. 
You can find more details about it in the corresponding section of my proposal https://docs.google.com/document/d/1xh7j4RtipN1XolYdNu4edhxNWEcfqu2C8UVevi4ZQUg/edit?usp=sharing

maybe @chrizandr can better explain what we actually want

Mec-iS:
IMHO we need to consolidate the existing functionalities before getting into improving the developer experience. Please run a double-check (a checklist maybe) about what it is currently usable in hydrus and what it is in the spec. I feel we still have something else to do before starting with adding features to improve the developer experience.

chrizandr:
[...]

PR #456, opened 2020-03-09
FEAT: Add prototype of dynamic routes in hydrus

sameshl:
This is  a POC of having dynamic routes in hydrus. For the prototype I am showing only how a custom **before** request function can be added for a specific route,  this can easily be extended to for **after** request function also. I have used an approach with decorators.

I am exposing a decorator called `custom_before_request` which will take two arguments:
1.`path`  - url route at which the user wants to execute the custom before request code
2.`method` - the HTTP verb at which the user to execute the custom code at  the specified url path

After this, when the decorator is applied on a user defined function, I save these three parameters(`path`, `method`, and the `function`  on which the decorator is applied) in a global double/nested dict called `before_request_funcs` in the form `{'path1':{'method1': function_to_be_called_at_this_path_and_method}}` .

In the hydrus backend I have declared a function called `before_request_callback` (which is decorated by flask's `before_request`  decorator) which checks the global `before_request_funcs` dict for any function to be called by checking the path and method of the current request and then calls the appropriate function.

Let me know what you think of this implementation and how it could be improved.
PS: 
* I have not included code for error handling, etc as I thought this was just a POC.
* I have tested it and it works locally.
* The functions have been defined in `app.py` and not been imported from any other file is to avoid the circular dependency issues in flask at the prototype stage.
I will fix this issue once the prototype is accepted

related to #404

#### To test the feature locally:
* activate your virtualenv
* `python3 app.py`
* Now, send a **GET** request to `http://localhost:8080/api/MessageCollection` endpoint and observe the console output in your terminal.
* You could also try sending a **PUT** request to `http://localhost:8080/api/MessageCollection` endpoint with appropriate body such as 
```
{
"MessageString": "Test string",
"@type": "Message"
}
```
and observe the console output.

### Checklist

### DIFF
--- a/hydrus/app.py
+++ b/hydrus/app.py
@@ -1,9 +1,14 @@
 """Main route for the application"""
 
 import logging
+import sys
+from os.path import dirname, abspath
+# insert the ./app.py file path in the PYTHONPATH variable for imports to work
+sys.path.insert(0, dirname(dirname(abspath(__file__))))
 
 from sqlalchemy import create_engine
-
+from flask import request
+from collections import defaultdict
 from gevent.pywsgi import WSGIServer
 from sqlalchemy.orm import sessionmaker
 
@@ -42,8 +47,8 @@ properties = doc_parse.get_all_properties(classes)
 doc_parse.insert_classes(classes, session)
 doc_parse.insert_properties(properties, session)
 
-AUTH = True
-TOKEN = True
+AUTH = False
+TOKEN = False
 
 if AUTH:
     try:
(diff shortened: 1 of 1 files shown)


Mec-iS:
There are PEP-8 warnings. Please [check here](https://travis-ci.com/HTTP-APIs/hydrus/jobs/296053259)

sameshl:
> There are PEP-8 warnings. Please [check here](https://travis-ci.com/HTTP-APIs/hydrus/jobs/296053259)

@me

vddesai1871:
How it handles GET request with query parameters?

sameshl:
> How it handles GET request with query parameters?

I have not implemented that yet. I was just waiting for general approval on my approach.
I will implement GET request with query parameters by tonight if the approach seems fine.

sameshl:
@vddesai1871  I just realized that the above implementation could also handle GET request with query parameters. The query parameters can be accessed by the user with `flask.request.args` and the user can take appropriate action with the query_parameters

PR #459, opened 2020-03-15
Refactor hydrus/resources.py

sameshl:
related to  #424 

### Checklist

### Description

Refactor `hydrus/resources.py`. This file had very poor readability and maintainability as it had all the logic for response generation and a lot of the code was repeating. I abstracted away the logic of response creation to lower level functions instead of keeping it in resources.py, thus,  increasing readability. Now resources.py only imports required helper functions(handles) instead of importing a lot of utility functions and exception classes. 

This is a big PR so these are some points on the changes I made so it becomes easier to read through the code:
* Replace the repeating authentication of request on every http method with decorator, like a *middleware*.
 * move validation logic for an object before it is inserted/updated
into separate function called `validate_object` in helpers.py
* reduce the function call of `get_doc` while generating the collections
and parsed classes of an api_doc and moved it to a separate function
* `get_context` function abstacts away logic of dynamically generating response.
* `error_response` function generates responses for errors during operations
* `send_update` function is handler for sending synchronization update to all
connected clients
* create `hydrus/item_collection_helpers.py` to handle the response generation
of GET, PUT, POST and DELETE operations on Item **Collection** classes.
* create `hydrus/items_helpers.py` to handle the response generation
of PUT and DELETE operation on **multiple** items.

### DIFF
--- a/hydrus/auth.py
+++ b/hydrus/auth.py
@@ -1,4 +1,5 @@
 from typing import Union
+from functools import wraps
 
 from flask import jsonify, Response, request
 from hydra_python_core.doc_writer import HydraError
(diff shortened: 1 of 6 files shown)


sameshl:
@xadahiya @chrizandr Could you please review this? Thanks!

PR #461, opened 2020-03-18
Refactor hydrus

sameshl:
related to #424 

### Checklist

### Description

* refactor error responses with `error_response` function
* refactor validation logic of objects while inserting / updating using `validate_object` function
* replace  `.format(...)` strings with Python 3.6+'s "f-strings" in the whole hydrus codebase. Not only are f-strings more readable, more concise, and less prone to error than other ways of formatting, but they are also faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,8 @@
 language: python
 python:
-  - "3.5"
-  - "3.5-dev"  # 3.5 development branch
   - "3.6"
-  - "3.6-dev"  # 3.6 development branch
-  - "3.7-dev"
+  - "3.7"
+  - "3.8"
 install:
   - pip install -r requirements.txt --no-cache
 script: python -m unittest discover
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 python:
   - "3.6"
   - "3.7"
-  - "3.8"
 install:
   - pip install -r requirements.txt --no-cache
 script: python -m unittest discover
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 language: python
 python:
   - "3.6"
+  - "3.6-dev"  # 3.6 development branch
   - "3.7"
+  - "3.7-dev" # 3.7 development branch
 install:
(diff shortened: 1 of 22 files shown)


sameshl:
@xadahiya The build is failing for python version < 3.6, where f-strings were not introduced. Do you still want to keep the changes? I thought f strings would be a great improvement to the codebase in terms of readability/maintainability.

Mec-iS:
> @xadahiya The build is failing for python version < 3.6, where f-strings were not introduced. Do you still want to keep the changes? I thought f strings would be a great improvement to the codebase in terms of readability/maintainability.

I am also starting to think that we may drop compatibility of Python 3.5, as for now everybody is on the move to 3.6 and we can easily do it at this stage of adoption

sameshl:
> I am also starting to think that we may drop compatibility of Python 3.5, as for now everybody is on the move to 3.6 and we can easily do it at this stage of adoption

Yes, exactly.

xadahiya:
The build is failing. @sameshl, please check

sameshl:
@xadahiya As mentioned earlier the build is failing only for python < 3.6 as I have used f strings while refactoring.

xadahiya:
Okay, I think there's no harm in dropping support for python < 3.6. Can you update travis config and README and related docs? Then we can merge

sameshl:
> Okay, I think there's no harm in dropping support for python < 3.6. Can you update travis config and README and related docs? Then we can merge

Yes. Will do

.travis.yml
```diff
@@ -1,10 +1,8 @@
 language: python
 python:
-  - "3.5"
-  - "3.5-dev"  # 3.5 development branch
   - "3.6"
-  - "3.6-dev"  # 3.6 development branch
```
xadahiya: Why did you remove 3.6-dev and 3.7-dev?

sameshl: Ah. I thought we would not need to test for the dev branches if we are building against python 3.8.

sameshl: Will revert the change.

sameshl:
@xadahiya I have made the required changes. Could you have a look?

PR #478, opened 2020-05-29
Refactor crud.py

sameshl:
Related to #424 

### Checklist

### Description

Refactor parts of `hydrus/data/curd.py`.

### DIFF
--- a/hydrus/data/crud.py
+++ b/hydrus/data/crud.py
@@ -199,99 +199,27 @@ def insert_multiple(objects_: List[Dict[str,
             valid/defined RDFClass but is not a dictionary neither an Abstract Property
 
     """
-    # instance list to store instances
-    instance_list = list()
-    triples_list = list()
-    properties_list = list()
-    instances = list()
     id_list = id_.split(',')
+    # list to hold all the ids of inserted objects
     instance_id_list = list()
 
-    # the number of objects would be the same as number of instances
+    link_props_of_object_ = dict()
+    id_of_object_ = None
     for index in range(len(objects_)):
-        type_ = objects_[index]["@type"]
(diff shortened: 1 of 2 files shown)


hydrus/data/crud.py
```diff
@@ -254,11 +209,8 @@ def insert_multiple(objects_: List[Dict[str,
```
chrizandr, line 209: Can we not reuse insert here as is? What is preventing us to use insert in insert_mutliple? All these checks are already in insert, maybe we should reuse that function?

hydrus/data/crud.py
```diff
@@ -410,30 +346,18 @@ def delete_multiple(
...
-        data_III += session.query(triples).filter(
-            triples.GraphIII.subject == index).all()
+        instance = get_instance_before_delete(session, index, type_)
+        instances.append(instance)
+        data_IAC_index, data_III_index, data_IIT_index = get_data_iac_iii_iit(session, index)
+        data_IIT += data_IAC_index
+        data_IAC += data_IAC_index
+        data_III += data_III_index
 
     data = data_III + data_IIT + data_IAC
```
chrizandr, line 362: Same here, these checks are already in delete. Maybe we can reuse the function with a few modifications.

hydrus/data/crud.py
sameshl: Yes, makes sense. I will make this change

hydrus/data/crud.py
sameshl: Yes, I will do this

sameshl:
@chrizandr Made the requested changes

xadahiya:
@chrizandr Please confirm the necessary changes are done, then we can merge

Issue #412, opened 2019-07-19
[Discussion] Research better multiple-table architecture (objetive 2021)


xadahiya:
Research and discuss a multiple-table database architecture where different type of resources are stored in different tables to improve scalability and efficiency.

Guttz:
To make the discussion easier I figured it would be good to deploy an instance of the DB with an IDE.
Available at: https://sqliteonline.com/#fiddle-5d31c0c75d188js6jya4czcr @Mec-iS

![image](https://user-images.githubusercontent.com/11739405/61537518-bf29df80-aa0d-11e9-9961-b56c1f3c482f.png)

Mec-iS:
> Available at: https://sqliteonline.com/#fiddle-5d31c0c75d188js6jya4czcr

This should go in the doc (:

Mec-iS:
As discussed we may be moving to TTL mappings [R2RDF](https://www.w3.org/TR/r2rml/) in 2021. Further discussion to come.

PR #479, opened 2020-06-10
Make db tables for each resource in apidoc

sameshl:
Related #412 

### Checklist

### Description

I have wrote POC code for making the database tables from parsing the API Doc.
This makes a table for each `hydra:Resource` (Classes and Collections both).
At the moment, the database schema is just a table for each resource.
The columns in each table are its all the properties in that classes / collection's `supportedProperty`.
Right now, I have just defined all these columns using the `STRING` class from sqlachemy.
There is also, a `id` column in each table acting as a primary key.

The database schema looks like, 
image

(I have attached the schema in the form of the SQL commands used to make that table, will update it with a proper visual schema. I did not find a tool to convert my `.db` file, to a visual schema.)

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -8,7 +8,7 @@ from hydrus.utils import (set_session, set_doc, set_hydrus_server_url,
                           set_page_size, set_pagination)
 from hydrus.data import doc_parse
 from hydra_python_core import doc_maker
-from hydrus.data.db_models import Base
+from hydrus.data.db_models import Base, create_database_tables
 from hydrus.data.user import add_user
 from hydrus.data.exceptions import UserExists
 from hydrus.data.stale_records_cleanup import remove_stale_modification_records
@@ -90,7 +90,6 @@ def serve(adduser: tuple, api: str, auth: bool, dburl: str, pagination: bool,
     # See http://docs.sqlalchemy.org/en/rel_1_0/core/engines.html for more info
     # DB_URL = 'sqlite:///database.db'
     DB_URL = dburl
-
     # Define the server URL, this is what will be displayed on the Doc
     HYDRUS_SERVER_URL = f"{serverurl}:{str(port)}/"
 
@@ -103,10 +102,10 @@ def serve(adduser: tuple, api: str, auth: bool, dburl: str, pagination: bool,
     engine = create_engine(DB_URL, connect_args={'check_same_thread': False})
     # Add the required Models to the database
 
-    if use_db is False:
-        click.echo("Creating models")
-        Base.metadata.drop_all(engine)
-        Base.metadata.create_all(engine)
+    # if use_db is False:
+    #     click.echo("Creating models")
+    #     Base.metadata.drop_all(engine)
+    #     Base.metadata.create_all(engine)
     # Define the Hydra API Documentation
     # NOTE: You can use your own API Documentation and create a HydraDoc object
     # using doc_maker or you may create your own HydraDoc Documentation using
@@ -167,12 +166,15 @@ def serve(adduser: tuple, api: str, auth: bool, dburl: str, pagination: bool,
 
     # Get all the properties from the classes
     properties = doc_parse.get_all_properties(classes)
-
     # Insert them into the database
     if use_db is False:
+        Base.metadata.drop_all(engine)
         click.echo("Adding Classes and Properties")
-        doc_parse.insert_classes(classes, session)
-        doc_parse.insert_properties(properties, session)
+        create_database_tables(classes)
+        click.echo("Creating models")
+        Base.metadata.create_all(engine)
+        # doc_parse.insert_classes(classes, session)
+        # doc_parse.insert_properties(properties, session)
 
     # Add authorized users and pass if they already exist
     click.echo("Adding authorized users")
--- a/cli.py
+++ b/cli.py
@@ -100,12 +100,6 @@ def serve(adduser: tuple, api: str, auth: bool, dburl: str, pagination: bool,
     click.echo("Setting up the database")
     # Create a connection to the database you want to use
     engine = create_engine(DB_URL, connect_args={'check_same_thread': False})
-    # Add the required Models to the database
-
-    # if use_db is False:
-    #     click.echo("Creating models")
-    #     Base.metadata.drop_all(engine)
-    #     Base.metadata.create_all(engine)
     # Define the Hydra API Documentation
     # NOTE: You can use your own API Documentation and create a HydraDoc object
     # using doc_maker or you may create your own HydraDoc Documentation using
@@ -163,9 +157,6 @@ def serve(adduser: tuple, api: str, auth: bool, dburl: str, pagination: bool,
     # You can also pass dictionary defined in
     # hydra_python_core/doc_writer_sample_output.py
     classes = doc_parse.get_classes(apidoc.generate())
-
-    # Get all the properties from the classes
-    properties = doc_parse.get_all_properties(classes)
     # Insert them into the database
     if use_db is False:
         Base.metadata.drop_all(engine)
@@ -173,8 +164,6 @@ def serve(adduser: tuple, api: str, auth: bool, dburl: str, pagination: bool,
         create_database_tables(classes)
         click.echo("Creating models")
         Base.metadata.create_all(engine)
-        # doc_parse.insert_classes(classes, session)
-        # doc_parse.insert_properties(properties, session)
 
     # Add authorized users and pass if they already exist
     click.echo("Adding authorized users")
--- a/hydrus/data/crud.py
+++ b/hydrus/data/crud.py
@@ -240,12 +240,6 @@ def get_collection(API_NAME: str,
         ClassNotFound: If `type_` does not represent a valid/defined RDFClass.
 
     """
-    collection_template = {
-        "@id": f"/{API_NAME}/{path}/",
-        "@context": None,
-        "@type": f"{type_}Collection",
-        "members": list()
-    }  # type: Dict[str, Any]
     database_search_params = copy.deepcopy(search_params)
     pagination_parameters = ["page", "pageIndex", "limit", "offset"]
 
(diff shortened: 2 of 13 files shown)


chrizandr:
The next step in my opinion would be to get CRUD operations to work with this new schema. And see how we can store the Graph. We can use the existing mechanism for graph storage with some modificaitons if needed. Also keep in mind the linking of the columns to their respective tables. I would suggest using the core library to parse the API doc and figure out which properties will require tables of their own. And linking them to the class.

sameshl:
Yes. I am done with the CRUD operations part.
For the graph, we should be able to use the existing mechanism as you suggested with some minor changes.
For now, though, I am looking into the linking of columns

sameshl:
@chrizandr @xadahiya The PR is ready for review.
The build is failing atm as it is running all tests in `/hydrus/tests/` in one session. And due to some issue due to the `Base` import [here](https://github.com/HTTP-APIs/hydrus/blob/6e39e0518d6536dce6921ebb24cd4a76bf1f5f72/hydrus/tests/conftest.py#L14), it's using the same Base instance for all the different database schemas. That is causing this error.
```
sqlalchemy.exc.InvalidRequestError: Table 'dummyClass' is already defined for this MetaData instance.  Specify 'extend_existing=True' to redefine options and columns on an existing Table object.
```

**I ran all the tests files individually in the folder `/hydrus/tests` and  `/tests` and all of them are passing**.

I am working on fixing this import issue atm. But the code can be merged as all the tests are passing individually.

They are failing in the build as the imports in `hydrus/tests/conftest.py` are only done once in the whole build session and due to issue with `Base` import, the build is failing.

sameshl:
@chrizandr @xadahiya I have fixed the bug in tests. All tests are passing now. Also, added doc strings and type annotations.
PR can be merged now

hydrus/data/resource_based_classes.py
```diff
@@ -0,0 +1,209 @@
...
+    database_class = Resource.all_database_classes.get(type_, None)
+    if database_class is None:
+        raise ClassNotFound(type_)
+    return database_class
+
+
+def insert_object(object_: Dict[str, Any], session: scoped_session) -> str:
+    """
+    Insert the object in the database
+    :param object_: Dict containing object properties
```
chrizandr, line 47: From what I can see the `object_` dictionary is the same as the one passed in the request right? In that case the property resource class name and the key in `object_` may not be same.

For example in `Drone` we can have `DroneState` object mapped to the "state" key in the request, but the table name would be DroneState. Where are you mapping the keys to the actual resource classes?

hydrus/data/resource_based_classes.py
```diff
@@ -0,0 +1,209 @@
...
+            session.query(database_class)
+            .filter(database_class.id == id_)
+            .one()
+        ).__dict__
+    except NoResultFound:
+        raise InstanceNotFound(type_=type_, id_=id_)
+    # Remove the unnecessary keys from the object retrieved from database
+    object_template = copy.deepcopy(object_)
+    object_template.pop("_sa_instance_state")
+    object_template.pop("id")
```
chrizandr, line 116: Why remove ID?

hydrus/data/resource_based_classes.py
```diff
@@ -0,0 +1,209 @@
...
+            session.query(database_class)
+            .filter(database_class.id == id_)
+            .one()
+        )
+    except NoResultFound:
+        raise InstanceNotFound(type_=type_, id_=id_)
+    session.delete(object_)
+    try:
+        session.commit()
+    except Exception:
```
chrizandr, line 141: It's not a good idea to use generic exceptions when using an external library. Try and figure out what might go wrong in SQLAlchemy and catch those specific exceptions.

hydrus/data/resource_based_classes.py
```diff
@@ -0,0 +1,209 @@
...
+    # Keep the object as fail safe
+    old_object = get_object(query_info, session)
+    # Delete the old object
+    delete_object(query_info, session)
+    id_ = query_info["id_"]
+    # Try inserting new object
+    try:
+        object_["id"] = id_
+        d = insert_object(object_, session)
+    except Exception as e:
```
chrizandr, line 166: Here generic exceptions are fine because we know that insert object will throw only certain exceptions.

hydrus/data/resource_based_classes.py
```diff
@@ -0,0 +1,209 @@
...
+    :param search_params: Query parameters
+    :param type_: @type of object to be deleted
+    :return: filtered instances
+    """
+    database_class = get_database_class(type_)
+    try:
+        filtered_instances = (
+            session.query(database_class).filter_by(**search_params).all()
+        )
+    except Exception as e:
```
chrizandr, line 189: Again don;t use generic exceptions with SQLAlchemy.

hydrus/item_collection_helpers.py
```diff
@@ -14,7 +14,10 @@
     InstanceNotFound,
     PageNotFound,
     InvalidSearchParameter,
-    OffsetOutOfRange)
+    OffsetOutOfRange,
+    DatabaseConstraintError,
+    PropertyNotGiven
+)
```
chrizandr, line 20: See #481 for line 89

hydrus/tests/unit/test_crud.py
```diff
@@ -33,31 +33,16 @@ def test_crud_get_returns_correct_object(drone_doc_collection_classes, drone_doc
 
 def test_get_for_nested_obj(drone_doc_collection_classes, drone_doc, session, constants):
     """Test CRUD get operation for object that can contain other objects."""
-    API_NAME = constants['API_NAME']
```
chrizandr, line 36: Do we need to add any new tests to the crud operations for the new schema? You added a few new exceptions, it might be a good idea to add tests for those cases.

hydrus/tests/conftest.py
```diff
@@ -46,9 +46,13 @@ def gen_dummy_object(class_title, doc):
     for class_path in doc.parsed_classes:
         if class_title == doc.parsed_classes[class_path]["class"].title:
             for prop in doc.parsed_classes[class_path]["class"].supportedProperty:
-                if isinstance(prop.prop, HydraLink) or prop.write is False:
+                if prop.write is False:
                     continue
-                if "vocab:" in prop.prop:
+                if isinstance(prop.prop, HydraLink):
```
chrizandr, line 51: Again treat hydra:link as just links, not as resources.

hydrus/tests/conftest.py
```diff
@@ -46,9 +46,13 @@ def gen_dummy_object(class_title, doc):
...
+                    prop_class = prop.prop.range.replace("vocab:", "")
+                    object_[prop.title] = gen_dummy_object(prop_class, doc)
+                    pass
+                elif "vocab:" in prop.prop:
```
chrizandr, line 55: Here also we need to consider the third case when vocab is expanded.

hydrus/data/crud_helpers.py
```diff
@@ -1,48 +1,15 @@
...
 from sqlalchemy.orm.scoping import scoped_session
-from sqlalchemy.orm import with_polymorphic
 
 from typing import Dict, Any, Tuple
 
 from sqlalchemy.orm.exc import NoResultFound
-from hydrus.data.db_models import (Graph, BaseProperty, RDFClass, Instance,
-                                   Terminal)
 from hydrus.data.exceptions import (
-    ClassNotFound,
```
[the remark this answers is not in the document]
sameshl: They were just removed from this file. In the new schema, we are still using it but they are being used in `/data/resource_based_classes.py` file like here https://github.com/HTTP-APIs/hydrus/pull/479/files#diff-fa37f0d4c7f3c7a8446f20c2ed5ccc4bR40

cli.py
```diff
@@ -103,10 +102,10 @@ def serve(adduser: tuple, api: str, auth: bool, dburl: str, pagination: bool,
     engine = create_engine(DB_URL, connect_args={'check_same_thread': False})
     # Add the required Models to the database
 
-    if use_db is False:
-        click.echo("Creating models")
-        Base.metadata.drop_all(engine)
-        Base.metadata.create_all(engine)
+    # if use_db is False:
```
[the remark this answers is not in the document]
sameshl: Yes.

hydrus/data/resource_based_classes.py
sameshl: Yes, the `object_` dictionary is the same as the one passed in the request.
From what I understood, your doubt is that in the request we would pass the id of a **State** instance in the key called `DroneState`. But the table in the database is called `State`, so where am I mapping the `DroneState` key to the `State` table instance.

That I am doing in [here](https://github.com/HTTP-APIs/hydrus/blob/27db73939f2fe8a8d08a8e3c2096876247dcf7d0/hydrus/data/resource_based_classes.py#L63). While creating the tables itself, I am adding a dict in the `.info` attribute of any foreign key through which I can extract the column name, which then helps in extracting the correct value from the `object_` dict.

Does that make it clear?

hydrus/data/resource_based_classes.py
sameshl: Two reasons:
1) The `id` for the object is actually inserted in `crud.py` [here](https://github.com/HTTP-APIs/hydrus/blob/27db73939f2fe8a8d08a8e3c2096876247dcf7d0/hydrus/data/crud.py#L84).
2) In the response we need to have the `id` field as `@id` key, so removing the `id` key.

hydrus/data/resource_based_classes.py
sameshl: Understood :+1:

hydrus/item_collection_helpers.py
sameshl: Yes, I understood the problem we have atm with notation of Collection Resources. I will make a new PR to fix this issue

hydrus/data/crud.py
```diff
@@ -80,12 +75,11 @@ def get(id_: str, type_: str, api_name: str, session: scoped_session,
...
-    rdf_class = get_rdf_class(session, type_)
-    object_template = add_prop_name_to_object(session, id_, object_template, rdf_class)
-    object_template["@type"] = rdf_class.name
+    query_info = {
+        "@type": type_,
+        "id_": id_
+    }
+    object_template = get_object(query_info, session)
     if path is not None:
         object_template["@id"] = f"/{api_name}/{path}Collection/{id_}"
```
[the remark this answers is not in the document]
sameshl: Yes, we should ideally be extracting the name of the collection from the API documentation itself. I will make this change in a new PR as that is a big change in itself

hydrus/data/crud.py
```diff
@@ -343,7 +246,15 @@ def get_collection(API_NAME: str,
...
+    pagination_parameters = ["page", "pageIndex", "limit", "offset"]
+
+    # remove pagination params before filtering in the database
+    for param in database_search_params.copy():
+        if param in pagination_parameters:
+            database_search_params.pop(param)
+
+    filtered_instances = get_all_filtered_instances(session, database_search_params, type_)
     result_length = len(filtered_instances)
     try:
```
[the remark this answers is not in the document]
sameshl: Yes, thats a nice idea.

hydrus/data/crud_helpers.py
```diff
@@ -63,7 +30,7 @@ def recreate_iri(API_NAME: str, path: str, search_params: Dict[str, Any]) -> str
 
 
 def parse_search_params(search_params: Dict[str, Any],
-                        properties: BaseProperty,
+                        properties,
```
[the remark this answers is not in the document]
sameshl: Ahh, I don't know how I missed that. Fixed that

hydrus/tests/conftest.py
sameshl: So I should treat `hydra:link` as just a normal column in our new schema right? Because as you had said that the burder for `hydra:link` falls on the client then

hydrus/data/db_models.py
```diff
@@ -1,242 +1,152 @@
...
+                String,
+                default=lambda: str(uuid.uuid4()),
+                unique=True,
+                primary_key=True,
+            ),
+        }
+        for supported_property in self.supported_properties:
+            title = supported_property["title"]
+            link = supported_property["property"]
+            if type(link) is not dict:
```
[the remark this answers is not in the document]
sameshl: I had got some issues regarding this. Will open a issue for expansion of the `vocab` keyword

sameshl:
> I see no changes to the main hydrus code. I would guess that the new exceptions you created atleast need to be caught there. Please check and confirm

I have made changes to main hydrus code too, where I am catching the exceptions. For eg [here](https://github.com/HTTP-APIs/hydrus/blob/f9d25b67b0b353d778ade1c6defc3feaae81ad69/hydrus/item_collection_helpers.py#L139)

hydrus/data/resource_based_classes.py
sameshl: Yes, I understand. I will not do that next time :sweat_smile:

hydrus/tests/unit/test_crud.py
sameshl: Yes, I added them

hydrus/data/resource_based_classes.py
sameshl: Yes. Fixed it

hydrus/data/db_models.py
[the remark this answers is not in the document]
sameshl: @chrizandr I have made an issue for the `vocab:` keyword expansion. We can discuss more on that there https://github.com/HTTP-APIs/hydrus/issues/482

sameshl:
@chrizandr I have made all the changes that you requested

hydrus/tests/unit/test_crud.py
```diff
@@ -265,3 +263,27 @@ def test_delete_multiple_id(drone_doc_collection_classes, drone_doc, session):
...
+                    nested_class = prop.prop.range.replace('vocab:', '')
+                else:
+                    nested_class = prop.prop.replace('vocab:', '')
+                continue
+    # remove the foreign key resource on purpose for testing
+    dummy_obj.pop(nested_class)
+    id_ = str(uuid.uuid4())
+    try:
+        insert_response = crud.insert(object_=dummy_obj, id_=id_, session=session)
+    except Exception as e:
```
chrizandr, line 286: Don't use generic exception anywhere in test. You should know exactly what error should occur.

chrizandr:
Good to merge once small change is done.

hydrus/tests/unit/test_crud.py
sameshl: Understood:+1:

sameshl:
@chrizandr Made the change :+1:

PR #483, opened 2020-07-10
Better discover collections

### LINKED ISSUE #481: Use of "*Collection" as default notation for Collection resources
### I'm submitting a 
- [ ] bug report.
- [x ] feature request.

### Current Behaviour:
In a lot of places hydrus assumes that the collection for a certain resource will be "[Classname]Collection". This has been added in a lot of places in the code as well.

For example in https://github.com/HTTP-APIs/hydrus/blob/master/hydrus/data/crud.py line 131. The same approach is also used in the modified schema. Collection endpoint and resource can use any "name" and we should not limit it to the current format.

### Expected Behaviour:
We should map collections to classes from the API documentation itself rather than from such naming schema.

sameshl:
Fixes #481 

### Checklist

### Description

In a lot of places, `hydrus` assumed that collections could only be "[classname]Collection".
Removed this explicit dependency on this notation.

### DIFF
--- a/hydrus/data/crud.py
+++ b/hydrus/data/crud.py
@@ -423,7 +423,7 @@ def pagination(filtered_instances, path, type_, API_NAME,
     collection_template = {
         "@id": f"/{API_NAME}/{path}/",
         "@context": None,
-        "@type": f"{type_}",
+        "@type": f"{path}",
         "members": list()
     }  # type: Dict[str, Any]
     result_length = len(filtered_instances)
--- a/hydrus/data/crud.py
+++ b/hydrus/data/crud.py
@@ -81,10 +81,7 @@ def get(id_: str, type_: str, api_name: str, session: scoped_session,
         "id_": id_
     }
     object_template = get_object(query_info, session)
-    if path is not None:
-        object_template["@id"] = f"/{api_name}/{path}Collection/{id_}"
-    else:
-        object_template["@id"] = f"/{api_name}/{type_}Collection/{id_}"
+    object_template["@id"] = f"/{api_name}/{path}/{id_}"
 
     return object_template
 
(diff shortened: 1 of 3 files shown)


hydrus/item_collection_helpers.py
```diff
@@ -85,7 +85,7 @@ def item_collection_get_response(path: str) -> Response:
             return error_response(error)
 
     # If endpoint and GET method is supported in the API and class is supported
-    if path in parsed_classes and f"{path}Collection" not in collections:
+    if path in parsed_classes:
```
chrizandr, line 88: Does parsed classes not contain collections anymore? Why is this function being called for non-collection classes?

sameshl:
> I see that you have removed the Collection from the @type string in pagination, but how we are ensuring that the Collection endpoint is there in path.

The collection endpoint will always be in the `path` as I had mentioned in the call. Also I ran this branch of `hydrus` against the agent and everything seems to work fine

hydrus/item_collection_helpers.py
sameshl: There was a slight confusion regarding that code. 

The function was being called for non-collection classes to accommodate for classes such as 'Area' which have the same endpoint(from the POV of `flask`) like any collection of the form `/serverapi/classname`.

hydrus/data/crud.py
```diff
@@ -81,10 +81,7 @@ def get(id_: str, type_: str, api_name: str, session: scoped_session,
         "id_": id_
     }
     object_template = get_object(query_info, session)
-    if path is not None:
-        object_template["@id"] = f"/{api_name}/{path}Collection/{id_}"
-    else:
-        object_template["@id"] = f"/{api_name}/{type_}Collection/{id_}"
+    object_template["@id"] = f"/{api_name}/{path}/{id_}"
```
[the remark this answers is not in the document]
sameshl: Yes, these changes are enough as the path will always contain the right endpoint(for a collection)

sameshl:
@chrizandr 
I have also made the required changes as you had requested. This PR can now be merged

sameshl:
@chrizandr @xadahiya This PR too can be merged.

Issue #487, opened 2020-07-22
[DISCUSSION] Database design for treating collections as a resource


sameshl:
Related to the issue https://github.com/HTTP-APIs/hydrus/issues/416, we need to plan the details of the necessary database changes needed to accommodate the new feature of 'Treating Collections as a Resource'.

From the last discussion, we have planned to have a new `collection_id` column for every collection table so as to distinguish which item belongs to which collection. We need new column as the values in this column could be repeated as a collection could have many items in it.

Apart from this column, the existing columns include `id`, which acts as the primary key for that table and `members` which would act as a link to the actual item in that item's table.

sameshl:
This issue can be closed as https://github.com/HTTP-APIs/hydrus/pull/488 has been merged.

PR #488, opened 2020-07-26
Treating Collection as a resource

### LINKED ISSUE #416: Treating collections as a resource
### I'm submitting a 
- [ ] bug report.
- [x] feature request.

### Current Behaviour:

I've been speculating for some time now that our treatment of collections as a single resource for every class is wrong. Nowhere in the spec is it mentioned that a collection is a set of all objects for a given class. It is mentioned that a collection is "a set of somehow related resources". This does not imply that a collection would be only limited to class types
### Expected Behaviour:

 I think the direction we must move in is for users to be able to define collections on their own. Each collection itself would have an `@id` since it is a subclass of `hydra:Resource`. We then use the collection endpoint to relate classes/properties.

We use Collections directly instead of having pseudo-Collections defined by parent_id or child_id. The main problem with both the above approaches is as you mentioned, identifying the property which contains the parent/child id. This is because schema.org/identifier or hydra:Link and even https://schema.org/ItemList are general definitions that can be used for any property and need not necessarily be for only parent/child ids. The advantage of using a collection is that we define a CommentCollection object, where we link all Comment objects for a particular issue. We then give that CommentCollection object as the property for the Issue. In terms of defining it in the API documentation, it would be something like:

```
{
        "@type": "SupportedProperty",
        "property": vocab:CommentCollection,
        "readonly": "false",
        "required": "true",
        "title": "Comments",
        "writeonly": "false"
}
```
The property would map to an instance of the CommanCollection class. This will be defined in the supportedProperty field for the definition of the Issue class. Like any other object, when we define the Issue object, we will add the appropriate instance of the CommandCollection to the Issue instance. So an object would be something like:

```
{
        "@type": "Issue",
        "@id": "/api/Issues/27",
        "Issue": "....",
        "Comments": "/api/CommentCollection/32",
        ....
}
```

And then `/api/CommentCollection/32` would have:

```
{
        "@type": "CommentCollection",
        "@id": "/api/CommentCollection/32",
        "members": [
                {
                        "@id": "/api/Comment/12",
                        "@type": "Comment"
                },
                {
                        "@id": "/api/Comment/18",
                        "@type": "Comment"
                },
                {
                        "@id": "/api/Comment/22",
                        "@type": "Comment"
                }
        ]

}
```
All this is part of the bigger implementation of treating collections as a resource. We allow hydrus to implement all operations: GET, PUT, POST, DELETE on all resource endpoints as well as on all collection endpoints. We then let users decide how they design their APIs and wether creation of resources happens at the collection endpoint or the resource endpoint is defined in the API Documentation itself. Nested collection would then be added to an object as any other resource.

NOTE: We should also allow the creation of both resources(Comment) as well as collections(CommentCollection) on the collection endpoint. We let the user decide which they want to use based on the API documentation. If the doc contains a `supportedOperation` to allow creation of resource using the collection endpoint then we create resources using that. Similarly for the case of resources. Hope this is clear.

sameshl:
Fixes #416 
Related to #487 

### Checklist

### Description

This PR adds supports for treating collection as a resource.

### Brief description of major changes to the endpoints after this PR is merged:
(Note: For below discussion, `CommandCollection` is a collection class, `Command` would be non-collection class / parsed class. The meaning of non-collection class would be any class which will not act as a 'collection' of items.)
1) **GET** on `/collection/`
##### Example endpoint: `/CommandCollection/`
This fetches the data from the 'CommandCollection' table.
##### Example response:
Will return a list of *collections*.
```json
{
    "@context": "/serverapi/contexts/CommandCollection.jsonld",
    "@id": "/serverapi/CommandCollection/",
    "@type": "CommandCollection",
    "hydra:totalItems": 3,
    "hydra:view": {
        "@id": "/serverapi/CommandCollection?page=1",
        "@type": "hydra:PartialCollectionView",
        "hydra:first": "/serverapi/CommandCollection?page=1",
        "hydra:last": "/serverapi/CommandCollection?page=1"
    },
    "members": [
        {
            "@id": "/serverapi/CommandCollection/7d2cc88f-388a-43f1-80fc-0c2184de4784",
            "@type": "CommandCollection"
        },
        {
            "@id": "/serverapi/CommandCollection/50ed3b68-9437-4c68-93d4-b67013b9d412",
            "@type": "CommandCollection"
        },
        {
            "@id": "/serverapi/CommandCollection/c2e10f88-d205-41e7-aa61-338afd64f657",
            "@type": "CommandCollection"
        }
    ]
}
```
2) **PUT** on `/collection/`
##### Example endpoint: `/CommandCollection/`
The request body should have the list of ids of class instances which would be grouped into a collection.
For eg,
```json
{
    "@type": "CommandCollection",
    "members": ["aaaaa",
                "bbbbb",
                "ccccc"]
}
```
In the above example, 'aaaaaa', 'bbbbb' and 'ccccc' are the `ids`(primary key) of the instances in `Command` table.
This adds data in the 'CommandCollection' table.
##### Example response:
```json
{
    "@type": "Status",
    "description": "Object with ID 50ed3b68-9437-4c68-93d4-b67013b9d412 successfully added",
    "statusCode": 201,
    "title": "Object successfully added."
}
```
**NOTE: The `id` returned is actually the value of the `collection_id` column in the table, not the primary key `id`.**
3) **GET** on `/collection/id`
##### Example endpoint: `/CommandCollection/50ed3b68-9437-4c68-93d4-b67013b9d412`
**Note: The `id` is corresponding to the `collection_id` column in the `CommandCollection` table, not the primary key `id`.**
##### Example response:
Will return all the members belonging to the collection with given `collection_id`.
```json
{
    "@context": "/serverapi/contexts/CommandCollection.jsonld",
    "@id": "/serverapi/CommandCollectionCollection/50ed3b68-9437-4c68-93d4-b67013b9d412",
    "@type": "CommandCollection",
    "members": [
        {
            "@id": "/serverapi/Command/aaaaaa",
            "@type": "Command"
        },
        {
            "@id": "/serverapi/Command/bbbbbb",
            "@type": "Command"
        },
        {
            "@id": "/serverapi/Command/cccccc",
            "@type": "Command"
        }
    ]
}
```
4) **POST** on `/collection/id`
##### Example endpoint: `/CommandCollection/50ed3b68-9437-4c68-93d4-b67013b9d412`
The request body should have the list of members to be updated for the given collection.
For eg,
```json
{
    "@type": "CommandCollection",
    "members": ["ddd",
                "eee",
                "fff"]
}
```
**Note: The `id` is corresponding to the `collection_id` column in the `CommandCollection` table, not the primary key `id`.**
##### Example response:
Will update all the members belonging to the collection with given `collection_id` with the members given in the request body.
```json
{
{
    "@type": "Status",
    "description": "Object with ID 50ed3b68-9437-4c68-93d4-b67013b9d412 successfully updated",
    "statusCode": 200,
    "title": "Object updated"
}
```
5) **DELETE** on `/collection/id`
##### Example endpoint: `/CommandCollection/50ed3b68-9437-4c68-93d4-b67013b9d412`
**Note: The `id` is corresponding to the `collection_id` column in the `CommandCollection` table, not the primary key `id`.**
##### Example response:
Will delete that collection from the table.
```json
{
    "@type": "Status",
    "description": "Object with ID 50ed3b68-9437-4c68-93d4-b67013b9d412 successfully deleted",
    "statusCode": 200,
    "title": "Object successfully deleted."
}
```
6) **GET**, **PUT**, **POST** and **DELETE** on any `/non-collection-class`
##### Example endpoint: `/Command/` or `/Area`
**NOTE: All of these will have the same behaviour as what would have happened before on `/CommandCollection/` endpoint**


sameshl:
@xadahiya @chrizandr Fixed all failing tests. The PR is ready to be merged now. You can have a look

hydrus/data/resource_based_classes.py
```diff
@@ -54,99 +56,143 @@ def insert_object(object_: Dict[str, Any], session: scoped_session) -> str:
...
-        # the name of the column through which this foreign key relationship
-        # is being established
-        fk_column = fk.info["column_name"]
+    if collection:
+        # if type_ is of a collection class
+        members = object_['members']
+        collection_id = id_ if id_ else str(uuid.uuid4())
+        for member in members:
+            # add all the members of that collection
+            inserted_object = database_class(members=member, collection_id=collection_id)
```
chrizandr, line 65: Isn't the collection supposed to have all members previously created before being added to the collection? Looks like they are being added dynamically?

hydrus/data/resource_based_classes.py
```diff
@@ -54,99 +56,143 @@ def insert_object(object_: Dict[str, Any], session: scoped_session) -> str:
...
-    object_template = copy.deepcopy(object_)
-    object_template.pop("_sa_instance_state")
-    object_template.pop("id")
-    object_template["@type"] = query_info["@type"]
-    return object_template
-
-
-def delete_object(query_info: Dict[str, str], session: scoped_session) -> None:
+    if collection:
+        objects = (
```
chrizandr, line 125: Better to move these things to separate functions and call them here instead of having long if else branches.

hydrus/data/resource_based_classes.py
sameshl: Yes, the collection is supposed to have all members previously created before being added to the collection.

Here, the `database_class` object refers to the collection table, not the instance table. So that object is for the collection table

sameshl:
@chrizandr Added tests for collection as a resource operations

hydrus/data/resource_based_classes.py
sameshl: Yes, will do

PR #490, opened 2020-08-09
Remove hardcoded dependencies on vocab keyword

### LINKED ISSUE #482: hydrus hardcoded with 'vocab:' to check foreign key relationship
### I'm submitting a 
- [x] bug report.
- [ ] feature request.

### Current Behaviour:

At the moment, `hydrus` has been hardcoded with `vocab:` while it tries to identify foreign key relationship. For eg, [here](https://github.com/HTTP-APIs/hydrus/blob/27db73939f2fe8a8d08a8e3c2096876247dcf7d0/hydrus/data/db_models.py#L90).

@chrizandr  suggested we also include the case where we expand the `vocab:` keyword and also check that for identifying foreign key relationship. For eg, something like https://www.markus-lanthaler.com/hydra/event-api/vocab#Class1.

But the problem here is that just expanding `vocab:` won't solve the underlying issue for discovering all foreign key relationships correctly. For example, we could as well have something called `myownvocab:` in `@context` node of the Hydra API doc and then that could be used to reference something like `myownvocab:Class1`.

We might also extract this correctly after some work at parsing the API doc. But if suppose our context becomes nested, it will become difficult to parse.

### Expected Behaviour:

The ideal behaviour should be that, `hydrus` starts with an expanded API doc. We should not parse the API doc without expanding it beforehand. This change should then ideally take place it the `hydra_python_core` library, which generates the API doc object which `hydrus` uses for parsing data required of the API doc. Therefore the output of the `hydra_python_core` libraries `doc_maker` module should be an API doc which is expanded beforehand.

sameshl:
Fixes #482 

### Checklist

### Description

`hydrus` depended on the `vocab` keyword in a lot of places for essential operations like identifying foreign key relationships, etc.
This was not ideal behaviour.

After support for expanded API DOCs was added in `hydra-python-core` repo in PR https://github.com/HTTP-APIs/hydra-python-core/pull/42, the core library now returns an expanded APIDOC.

So, in this PR, added support for expanded API DOC.

**Note:** The build is failing currently due to some minor bugs in the core repo which @priyanshunayan is currently fixing.

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -156,7 +156,7 @@ def serve(adduser: tuple, api: str, auth: bool, dburl: str, pagination: bool,
     # Get all the classes from the doc
     # You can also pass dictionary defined in
     # hydra_python_core/doc_writer_sample_output.py
-    classes = doc_parse.get_classes(apidoc.generate())
+    classes = doc_parse.get_classes(apidoc)
     # Insert them into the database
     if use_db is False:
         Base.metadata.drop_all(engine)
--- a/cli.py
+++ b/cli.py
@@ -96,7 +96,6 @@ def serve(adduser: tuple, api: str, auth: bool, dburl: str, pagination: bool,
     # The name of the API or the EntryPoint, the api will be at
     # http://localhost/<API_NAME>
     API_NAME = api
-
     click.echo("Setting up the database")
     # Create a connection to the database you want to use
     engine = create_engine(DB_URL, connect_args={'check_same_thread': False})
@@ -105,7 +104,6 @@ def serve(adduser: tuple, api: str, auth: bool, dburl: str, pagination: bool,
     # using doc_maker or you may create your own HydraDoc Documentation using
     # doc_writer [see hydra_python_core/doc_writer_sample]
     click.echo("Creating the API Documentation")
-
     if hydradoc:
         # Getting hydradoc format
         # Currently supported formats [.jsonld, .py, .yaml]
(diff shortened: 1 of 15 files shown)


priyanshunayan:
@sameshl Can you confirm everything's working fine by using this version of core: https://github.com/priyanshunayan/hydra-python-core/tree/minor-fixes

I have opened a draft PR in the core repo too: https://github.com/HTTP-APIs/hydra-python-core/pull/48

sameshl:
@chrizandr @xadahiya This PR is completed and is ready to be merged. The build is failing because `hydrus` depends on the `master` branch of the `hydra-python-core` repo in which the changes like expansion of the API DOC have not been merged yet. They are just merged till the `develop` branch in the `hydra-python-core`.


Also, there were some bugs which were fixed in @priyanshunayan's fork of `hydra-python-core` in the `minor-fixes` branch. The relevant PR in the core is [this](https://github.com/HTTP-APIs/hydra-python-core/pull/48). Therefore, the `minor-fixes` branch has the latest code for the core repo.

**I ran the test suite of `hydrus` against the `minor-fixes` branch(in @priyanshunayan's fork) and all the tests are passing. So this PR is ready to be merged**

chrizandr:
Build still fails after merging into master. Please check.

sameshl:
The build is failing due to import error of `pyld` and possibly also `requests`. There are requirements of the core but I think they are not in its `requirements.txt`.

@priyanshunayan Could you make a PR for the same?

PR #491, opened 2020-08-11
Update sample docs

### LINKED ISSUE #482: hydrus hardcoded with 'vocab:' to check foreign key relationship

priyanshunayan:
Fixes #482 

### Checklist

### Description

According to the latest changes in the core library, this PR updates the sample docs used throughout and also the sample code used to generate the sample docs. 
### Change logs

- Updated Dones Doc
- Update sample doc by doc_writer
- code to generate doc by doc_writer

### DIFF
--- a/examples/drones/doc.py
+++ b/examples/drones/doc.py
@@ -1,6 +1,6 @@
 """
 Generated API Documentation for Server API using server_doc_gen.py."""
-
+ 
 doc = {
     "@context": {
         "ApiDocumentation": "hydra:ApiDocumentation",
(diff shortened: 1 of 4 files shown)


chrizandr:
Check why builds are failing

sameshl:
Yes, the build is failing for the same reason I pointed out [here](https://github.com/HTTP-APIs/hydrus/pull/490#issuecomment-671349378).

chrizandr:
Builds still failing after 490 merger

PR #493, opened 2020-08-18
Add support for multiple resource type collections

### LINKED ISSUE #489: Multiple resource type collections
### I'm submitting a 
- [ ] bug report.
- [x] feature request.

### Current Behaviour:

After PR https://github.com/HTTP-APIs/hydrus/pull/488/ gets merged, `hydrus` will have support for treating collections as a resource.

But in that implementation, `collections` are restricted to only a collection of a single type class.
### Expected Behaviour:

As described in the [spec](https://www.hydra-cg.com/spec/latest/core/#collections), collections are described as a 'set of somehow related resources'. This would also mean that a collection might also be a set of classes of different types.

We need to add support for this type of collections too.

sameshl:
Fixes #489

### Checklist

### Description

Now, collections support instances of different classes as members. :rocket:

##### Example usage after changes made
1) **PUT** on `/collection/`
The request body should have the list of `@id`s of class instances which would be grouped into a collection.
**NOTE: These instances should exist in the database before grouping them into a collection.**
For eg,
```json
{
  "@type": "LogEntryCollection",
  "members": [
    {
      "@id": "/serverapi/Drone/4ff14a9e-9cd0-4e8a-9c11-86ac9bec211f",
      "@type": "Drone"
    },
    {
      "@id": "/serverapi/LogEntry/aab38f9d-516a-4bb2-ae16-068c0c5345bd",
      "@type": "LogEntry"
    }
  ]
}
```
In the above example, a `LogEntry` instance with id(primary key) 'aab38f9d-516a-4bb2-ae16-068c0c5345bd' and a `Drone` instance with id(primary key) '4ff14a9e-9cd0-4e8a-9c11-86ac9bec211f' already exist in the database.
This adds data in the 'LogEntryCollection' table.
##### Example response:
```json
{
    "@type": "Status",
    "description": "Object with ID 50ed3b68-9437-4c68-93d4-b67013b9d412 successfully added",
    "statusCode": 201,
    "title": "Object successfully added."
}
```

2) **GET** on `/collection/id`
##### Example response:
Will return all the members belonging to the collection with given `collection_id`. The members returned are all `hydra:Links`.
```json
{
  "@context": "/serverapi/contexts/LogEntryCollection.jsonld",
  "@id": "/serverapi/LogEntryCollection/50ed3b68-9437-4c68-93d4-b67013b9d412",
  "@type": "LogEntryCollection",
  "members": [
    {
      "@id": "/serverapi/LogEntry/aab38f9d-516a-4bb2-ae16-068c0c5345bd",
      "@type": "hydra:Link"
    },
    {
      "@id": "/serverapi/Drone/4ff14a9e-9cd0-4e8a-9c11-86ac9bec211f",
      "@type": "hydra:Link"
    }
  ]
}
```

### DIFF
--- a/hydrus/helpers.py
+++ b/hydrus/helpers.py
@@ -224,17 +224,14 @@ def finalize_response(path: str, obj: Dict[str, Any]) -> Dict[str, Any]:
     collections, parsed_classes = get_collections_and_parsed_classes()
     expanded_base_url = DocUrl.doc_url
     if path in collections:
-        # path is of a collection class
-        collection = get_doc().collections[path]["collection"]
-        class_name = collection.manages["object"].split(expanded_base_url)[1]
-        collection_manages_class = get_doc().parsed_classes[class_name]["class"]
-        collection_manages_class_type = collection_manages_class.title
-        collection_manages_class_path = collection_manages_class.path
         members = list()
-        for member_id in obj["members"]:
+        for member in obj["members"]:
+            member_id = member[0]
+            member_type = member[1]
+            member_path = get_path_from_type(member_type)
             member = {
-                "@type": collection_manages_class_type,
-                "@id": f"/{get_api_name()}/{collection_manages_class_path}/{member_id}",
+                "@type": "hydra:Link",
+                "@id": f"/{get_api_name()}/{member_path}/{member_id}",
             }
             members.append(member)
         obj['members'] = members
@@ -500,3 +497,12 @@ def get_type_from_path(path: str) -> str:
         class_ = parsed_classes[class_name]['class']
         if path == class_.path:
             return class_.id_.split(expanded_base_url)[1]
+
+
+def get_path_from_type(type_: str) -> str:
+    _, parsed_classes = get_collections_and_parsed_classes()
+    expanded_base_url = DocUrl.doc_url
+    for class_name in parsed_classes:
+        class_ = parsed_classes[class_name]['class']
+        if type_ == class_.id_.split(expanded_base_url)[1]:
+            return class_.path
--- a/hydrus/helpers.py
+++ b/hydrus/helpers.py
@@ -506,3 +506,31 @@ def get_path_from_type(type_: str) -> str:
         class_ = parsed_classes[class_name]['class']
         if type_ == class_.id_.split(expanded_base_url)[1]:
             return class_.path
+
+
+def parse_collection_members(object_: dict) -> dict:
+    """Parse the members of a collection to make it easier
+    to insert in database.
+
+    :param object_: The body of the request having object members
+    :type object_: dict
+    :return: Object with parsed members
+    :rtype: dict
+    """
+    members = list()
+    for member in object_['members']:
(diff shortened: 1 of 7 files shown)


hydrus/helpers.py
```diff
@@ -491,3 +488,49 @@ def send_update(method: str, path: str):
...
+    :rtype: dict
+    """
+    members = list()
+    for member in object_['members']:
+        # eg member "/serverapi/LogEntry/7b51ac78-e917-4884-b52e-04e65d19b810"
+        member_components = member.split('/')
+        member_id = member_components[-1]
+        member_path = member_components[-2]
+        member_type = get_type_from_path(member_path)
+        if crud.item_exists(member_type, member_id, get_session()):
```
chrizandr, line 527: Where are we checking if the member is managed by the collection or not?

hydrus/tests/conftest.py
```diff
@@ -46,7 +48,16 @@ def gen_dummy_object(class_title, doc):
     expanded_base_url = DocUrl.doc_url
     for class_path in doc.collections:
         if class_title == doc.collections[class_path]["collection"].name:
-            members = [str(uuid.uuid4()) for _ in range(3)]
+            members = list()
+            for _ in range(3):
```
chrizandr, line 52: This test also needs to be modified. We should only create dummy objects for classes in the manages block of the collection. For any other type it should return an error. Add a separate test for the error case.

hydrus/helpers.py
```diff
@@ -224,17 +224,14 @@ def finalize_response(path: str, obj: Dict[str, Any]) -> Dict[str, Any]:
...
-        # path is of a collection class
-        collection = get_doc().collections[path]["collection"]
-        class_name = collection.manages["object"].split(expanded_base_url)[1]
-        collection_manages_class = get_doc().parsed_classes[class_name]["class"]
-        collection_manages_class_type = collection_manages_class.title
-        collection_manages_class_path = collection_manages_class.path
         members = list()
-        for member_id in obj["members"]:
+        for member in obj["members"]:
+            member_id = member[0]
```
chrizandr, line 229: This needs to be modified so that we can use the discussed format for the request.

hydrus/helpers.py
sameshl: Currently, we are **not** checking if the member is managed by the collection or not. I did this on purpose as on reading the [spec](https://www.hydra-cg.com/spec/latest/core/#manages-block), it was written that "a `manages` node to the collection instructs the **client**".

So, I assumed that the `manages` block is instruction for the client rather than the server and therefore the overhead of checking if the member is managed by the collection or not falls on the client.

Let me know if we need to check the manages block at the server also?

hydrus/helpers.py
sameshl: Yes. I am working on this right now

hydrus/tests/conftest.py
sameshl: Okay. Understood. Will make the change

chrizandr (approved):

PR #495, opened 2020-08-19
Add docs on database of hydrus

sameshl:
### Description

Add docs on `hydrus` explaining how `hydrus` goes about creating the database(and the server) when given an API Doc.


chrizandr (changes requested):
Can you add a sample API doc and then show the generated sturcture in the database using a generated ER diagram?

sameshl:
> Can you add a sample API doc and then show the generated sturcture in the database using a generated ER diagram?

Yes, sure. On it!

sameshl:
@chrizandr I have added an ER diagram for a sample API Doc :+1:

Issue #512, opened 2020-10-01
/serverapi/vocab should allow reaching fragments


Mec-iS:
### I'm submitting a 
- [x] bug report.

Two different related issues here:

### Current Behaviour:
1. `/serverapi/vocab#Area` returns the full vocabulary.
2. at `http://localhost:8080/serverapi/contexts/EntryPoint.jsonld` see: 
```{
"@context": {
"Area": {
"@id": "http://localhost:8080/serverapi/vocab**#EntryPoint/Area**",  <<<<<
"@type": "@id"
},
```

### Expected Behaviour:
1. `/serverapi/vocab#Area` should return the `Area` object (with context) only, not the full vocabulary, i.e.:
```
{
  "@context": { ... },
  "@id": "http://localhost:8080/serverapi/vocab#Area",
  "@type": "hydra:Class",
  "description": "Class for Area of Interest of the server",
  "supportedOperation": [
   ....
```

2. ids in `http://localhost:8080/serverapi/contexts/EntryPoint.jsonld` should link to `serverapi/vocab#` fragments like `@id": "http://localhost:8080/serverapi/**vocab#Area**"`

raoashish10:
@Mec-iS I have been trying to work out a solution for this issue but I haven't gotten far. I feel like the issue here is the use of '#'. Maybe the string ahead of serverapi/vocab gets ignored? I tried changing it to a '/' instead in the '@id' key but it gave me an error saying that no such link exists which kind of confuses me because I have provided the link in the '@id' key

Mec-iS:
the idea is to provide for the JSON-LD a feature that is availalble in HTML pages. For example if you write this link into you browser `https://en.wikipedia.org/wiki/Representational_state_transfer#Layered_system` you are redirected to the paragraph "Layered_system" inside the linked page. The same should happen for the JSON-LD, if a fragment is specified, the server should return only the object requested.

Here it is a way you can implement this:
1. create a Flask decorator that reads the path for the request (`Request` object, `from flask import request`); if a fragment name is present in the request (you can use a regular expression to match the fragment name or simply read the URI backward up to the `#`; first one would be more solid) tell the handler to return only the context plus the object specified by the fragment name
2. apply the decorator to the view that returns the JSON-LD

raoashish10:
@Mec-iS Understood! I will try implementing it and sending in a PR

raoashish10:
[...]

PR #518, opened 2020-10-12
Adding fragments as parameters when displaying vocab

raoashish10:
Regarding #512 

### Checklist

### Description

I understand that the links with fragment identifiers will still not be dereferenceable but as per my discussion with @Mec-iS, I have added the fragments as parameters for GET requests. It works for classes and collections as well. Let me know if I have made any mistakes or any changes that are needed. I have also refactorized my code and added a function in the helpers.py file.
Example: http://localhost:8080/serverapi/vocab?frag=LogEntryCollection

 
#### Added

- Added a check for parameters when sending a GET request to api/vocab in the resources.py file
- Added a helper function in helpers.py to check for the fragment and return the object

Output (http://localhost:8080/serverapi/vocab?frag=Drone):
```
{
    "@context": {
        "ApiDocumentation": "hydra:ApiDocumentation",
        "description": "hydra:description",
        "domain": {
            "@id": "rdfs:domain",
            "@type": "@id"
        },
        ...
    },
    "supportedClass": [
        {
            "@id": "http://localhost:8080/serverapi/vocab#Drone",
            "@type": "hydra:Class",
            "description": "Class for a drone",
            "supportedOperation": [
                {         ....
}
```


Mec-iS:
I don't think the name `frag` for the parameter is right.

Let's do a poll about how to call it:
1. `class` like in `?class=Drone`
2. `hydraclass` like in `?hydraclass=Drone` when the fragment is an HydraClass object; the parameter name has to change according to the type that has to be returned
3. `resource` like in `?resource=Drone`
4. ... please provide more options ... @xadahiya @chrizandr

I would go for 3. as it is the simplest

raoashish10:
@Mec-iS resource does seem more apt, I will change the parameter name as per other suggestions

Mec-iS:
yes please. wait to hear about what people thinks.

chrizandr:
I think resource is fine. We can move ahead with it.

Mec-iS:
Do we need tests for this feature?

hydrus/helpers.py
```diff
@@ -530,3 +530,21 @@ def parse_collection_members(object_: dict) -> dict:
...
+
+def get_fragments(resource: str) -> dict:
+    """Gets a fragment of the main hydra vocabulary.
+
+    :param resource: Fragment specified in the request parameters
+    :type resource: str
+    :return: Object referred to by the fragment
+    :rtype: dict
+    """
+    resourcedict = dict()
```
sameshl: I think it might be better to rename the parameter `resourcedict` to `resource_dict` for better readability and keeping in line with PEP8 standard

hydrus/helpers.py
```diff
@@ -530,3 +530,21 @@ def parse_collection_members(object_: dict) -> dict:
...
+    gen_doc = get_doc().generate()
+    resourcedict['@context'] = {i: gen_doc['@context'][i] for i in sorted(gen_doc['@context'])}
```
sameshl:
```
resourcedict['@context'] = {i: gen_doc['@context'][i] for i in sorted(gen_doc['@context'])}
```
Is there any particular reason we need to sort here(`sorted`)?

hydrus/helpers.py
```diff
@@ -530,3 +530,21 @@ def parse_collection_members(object_: dict) -> dict:
...
+    vocab_route = get_doc().doc_name
+    matching_string = f'{get_hydrus_server_url()}{get_api_name()}/{vocab_route}'+'#'+resource
+    res = next((i for i in gen_doc['supportedClass'] if i['@id'] == matching_string), None)
```
sameshl: What exactly are you trying to here? (I think this line could be rewritten in a more readable way, instead of having a compact one-liner)

sameshl (changes requested):

sameshl:
@raoashish10, Also as @Mec-iS mentioned, it might be a good idea to have tests for this feature. Could you also add them?

hydrus/helpers.py
raoashish10: So the context which was being returned, was jumbled (not sorted) and I think I had noticed sorted keys in contexts before. Is there no specification for sorting?? Then I can remove the sorted keyword

hydrus/helpers.py
raoashish10: Sure! I will make that change

hydrus/helpers.py
raoashish10: So the line 548, is for basically iterating through all of the id's in the apidoc and finding the resource that is specified. Since the id's have a fragment identifier in them, in line 547, I just appended the resource specified with the vocab URI and a # before the resource. 

The reason I used the next keyword and converted it into a compact one liner is because I have read that iterating using next is faster usually. Actually this can be converted into a simple for loop block, if that is more suited I will change the line.

raoashish10:
@sameshl @Mec-iS Sure! I will look into adding tests. I haven't really written tests for any code before so I might take some time but I will get it done soon

raoashish10:
@Mec-iS @sameshl I am running into a problem with one of the conditions of test. I noticed something that the HYDRUS_SERVER_URL is set to http://hydrus.com/ but when I try to print the server url in the resource.py function for getting fragments, it is printed as http://localhost/ . I can't seem to figure out why this is happening. Any help would be appreciated

hydrus/helpers.py
sameshl: afaik, I don't think there is any specification for the keys to be in any sorted order. You can remove the `sorted` keyword.

/cc @priyanshunayan @chrizandr

hydrus/helpers.py
sameshl: Okay, I see. I would recommend converting this into a simple for loop block for readability and maintainability purposes.

sameshl:
> `matching_string = f'{get_hydrus_server_url()}{get_api_name()}/{vocab_route}'+'#'+resource` should have had the value "http://hydrus.com/demoapi/vocab#" but instead it holds the value : "localhost/demoapi/vocab#"

I think the correct way to find the matching id is something like how its been done [here](https://github.com/HTTP-APIs/hydrus/blob/0d1ccaa8fd2512d815f32c627327d0ce4770c220/hydrus/helpers.py#L103-L106).

Mec-iS:
Just a note:
`matching_string = f'{get_hydrus_server_url()}{get_api_name()}/{vocab_route}'+'#'+resource` makes few sense as it is already using f-strings, so it should be:
`matching_string = f'{get_hydrus_server_url()}{get_api_name()}/{vocab_route}#{resource}`

raoashish10:
@sameshl Thank you for the suggestion! But I instead just used regex to match the string as @Mec-iS has suggested in the issue, tests are working properly. Let me know if there is any mistake as this is my first time writing a test

sameshl (approved):
LGTM :+1:

chrizandr:
I can see the changes in hydrus, but the API documentation and responses still give links of the format `vocab#Resource` right?

Shouldn't that be changed first? The `get` function you defined with the `resource` parameter will never be used unless someone manually types the URI for that resource. The format for URIs in the API documentation and the responses given by the server itself need to be changed to `vocab? resource=Resource`

raoashish10:
@chrizandr You're right! I can change the links returned by the server to vocab?resource = Resource but I have a doubt, this change will have to be made in the hydra_python_core repo right?

chrizandr:
Yes exactly.

raoashish10:
@chrizandr I will add a PR in the hydra_python_core repo to implement this. Is that alright?

chrizandr:
Yes, once that is merged we can go ahead with this

raoashish10:
@chrizandr I am stuck in college work at the moment but I will surely send in a PR for the link formats as soon as possible

Mec-iS:
@sameshl @priyanshunayan @chrizandr status of this PR?

chrizandr:
This is still incomplete. The correct changes in core haven't been made yet, we also need to make hydrus serve the content on those new URIs.

Mec-iS:
<s>required changes should now be present in core master. please confirm and merge @chrizandr</s>

PR HTTP-APIs/hydra-python-core#61 still open and in need of test cases

PR #522, opened 2020-10-16
Docker refactoring

Mec-iS:
This PR implement Poetry for Docker and get rid of `requirements.txt`. This will avoid dependencies conflicts from now on thanks to the lock file.

I am not 100% sure the `Dockerfile` is done correctly, please try `docker-compose up --build` locally and provide suggestions about how to make this safer and more stable.


chrizandr:
I think you need to do `pip install poetry` in `before_install`. See example here: https://gitmemory.com/issue/sdispater/poetry/366/488720509

.travis.yml
```diff
@@ -1,11 +1,10 @@
 language: python
 python:
-  - "3.6"
-  - "3.6-dev"  # 3.6 development branch
   - "3.7"
   - "3.7-dev" # 3.7 development branch
 install:
-  - pip install -r requirements.txt --no-cache
+  - pip install poetry
```
king-11, line 6: From the poetry docs, the recommended method to install poetry is not to use pip rather their own script

king-11:
@Mec-iS I can work on this if you aren't

Mec-iS:
> @Mec-iS I can work on this if you aren't

sorry for the late reply. You don't need to ask for permission. You can just follow the Github Flow and open a PR.

PR #556, opened 2021-02-11
Unique constraint for member ID in collections

### LINKED ISSUE #555: Avoid insertion of duplicate members in a collection
### I'm submitting a 
- [x] bug report.
- [ ] feature request.

### Current Behaviour:

Currently, a member with the same ID can be added multiple times in the same collection.
### Expected Behaviour:

Member ID in Collection database table should be unique and hydrus should raise and handle the integrity error everytime an object with existing ID is inserted in a collection. A response with 400 (Bad Request) code should be returned instead of a server error.

### Steps to reproduce:

- Make a PUT request to a Collection endpoint : `http://localhost:8080/serverapi/<collection_name>/<collection-id>`
- Insert an object again with same Object ID in JSON body using PUT request.

### Environment:
 
* python version
* pip version
* OS details

Yes

farazkhanfk7:
Fixes #555

### Checklist

### Description

- Previously,  a member with the same ID can be added multiple times in the same collection.
- Member ID in Collection database table is changed to a unique key and hydrus handles the integrity error and raises InstanceExists error if an object with existing ID is inserted in a collection. A response with 400 (Bad Request) code is returned.

```
{
    "@type": "Error",
    "description": "Instance of type Movie with ID c1b62361-e9c6-4185-b496-b74dff4fb8a4 already exists",
    "statusCode": 400,
    "title": "Instance already exists."
}
```

### DIFF
--- a/hydrus/data/db_models.py
+++ b/hydrus/data/db_models.py
@@ -16,6 +16,7 @@ from sqlalchemy import (
 )
 from sqlalchemy.engine import Engine
 from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.schema import UniqueConstraint
 from sqlalchemy.sql import func
 from hydra_python_core.doc_writer import DocUrl
 # from hydrus.settings import DB_URL
@@ -101,8 +102,6 @@ class Resource:
                 attr_dict[title] = Column(String)
 
         if "manages" in self.resource:
-            # if the class is a collection, then member id should be unique
-            attr_dict[title] = Column(String, unique=True)
             # if the class is a collection, add an extra column for
             # collection id
             attr_dict['collection_id'] = Column(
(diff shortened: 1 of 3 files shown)


Mec-iS:
Please add a unit test in the proper test module. we cannot accept any change without at least one test case. Follow the current convention for other tests.

farazkhanfk7:
Sure! I'll add the tests.

Mec-iS:
also, it seems to me that you are working on the `title` property, the primary `uuid` is used as s primary key and in the `id_` property.

@samesh please code review

farazkhanfk7:
Yes, because in a collection table `id_` will be generated by default and it'll serve as the primary uuid. Whereas, member(id) will be taken from `title` of supported_property but it should be unique as well(This isn't the case with a Class), so that no member can be added more than once.

sameshl (commented):
Good work @farazkhanfk7, LGTM :+1:

sameshl (approved):

PR #557, opened 2021-02-11
Fix test index URI assertion

Mec-iS:
minor change


farazkhanfk7:
These changes are present in develop branch but it isn't synced with master branch yet.

PR #571, opened 2021-05-06
API Root_path redirected to Root_path/api_name

### LINKED ISSUE #570: Serving the API on the Root_path
### I'm submitting a 
- [ ] bug report.
- [x] feature request.

### Current Behaviour:

we know that currently, we get a 404 at the root path (`http://localhost:8080/`).
### Expected Behaviour:

As discussed in #508, It is possible to have the entry point object served at the root domain.
But it should also be served at root_path even if the api_name is given.

yes

Purvanshsingh:
Fixes #570 

### Checklist

### Current behaviour

we get a 404 at the root path (`http://localhost:8080/`).
### New expected behaviour

I have tried to redirect the root_path to root_path/api_name in `app_factory.py` so that the API can be served by accessing the root_path.

rootpathfix


Purvanshsingh:
@Mec-iS can we go ahead with this approach of redirecting API to `root_path URL`, as we know that we can't serve it on root_path but can redirect it so that user can see the HATEOS payload directly by accessing the root.

Mec-iS (approved):

PR #575, opened 2021-05-14
Logging using after_request method & hydrus_logger added

### LINKED ISSUE #572: Adding detailed logging to hydrus
### I'm submitting a 
- [ ] bug report.
- [x] feature request.

### Current Behaviour:

Currently, we don't have any detailed logging system for hydrus. No logs are being generated.
### Expected Behaviour:

We can generate the log files which could help the users for getting the `Object_id` for any future manipulation 
and developers for debugging faster.

We will have a separate log folder that will be generated while running hydrus server.

```
Logs
|
|_ _ hydrus.log
```
Log file format

```
Logs/hydrus.log

run_time : 12-may-12:00:00
API_Doc fetched Successfully.
Context generated.
Terminated!
Database setup Error.
```

yes, I and @farazkhanfk7 are working on this and changing the API response structure [issue](https://github.com/HTTP-APIs/hydrus/issues/573) .

Purvanshsingh:
Fixes #572

### Checklist

### Current behaviour

We don't have any logging mechanism for hydrus.
### New expected behaviour

I have tried to add the things covered in [this discussion](https://github.com/HTTP-APIs/hydrus/issues/572), Opening this PR for more clear Suggestions.

1. I have added a `after_request` method where we are logging the `TimeStamp job_Id Method / iri` (resource_url is not fetched now) 
2.  Created `hydrus_logger.py` which provide us configured structlog logger object
3. `LogConfig` file for configuring the logger object.
4.  Logs Directory is not setup yet, `hydrus.log` will be generated in the same directory.

Screenshot from 2021-05-15 05-27-47

### DIFF
--- a/hydrus/app_factory.py
+++ b/hydrus/app_factory.py
@@ -24,7 +24,8 @@ def app_factory(api_name: str = "api", vocab_route: str = "vocab") -> Flask:
     @app.after_request
     def logging_request_response(response):
         logger = HydrusLogger().get_logger()
-        logger.info("{} {}/<iri>".format(uuid.uuid4(), request.method))
+        logger.info(" {} {}:{} Object_created_at : {}, status_code : {} "
+                    .format(uuid.uuid4(), request.method, request.path, response.location, response.status))
         return response
 
     # Redirecting root_path to root_path/api_name
--- a/hydrus/app_factory.py
+++ b/hydrus/app_factory.py
@@ -1,7 +1,8 @@
-from flask import Flask, redirect
+from flask import Flask, redirect, request
+import uuid
 from flask_cors import CORS
 from flask_restful import Api
-
+from hydrus_logger import HydrusLogger
 from hydrus.resources import (Index, Vocab, Contexts, Entrypoint,
                               ItemCollection, Item, Items, ItemMember)
 
@@ -19,6 +20,12 @@ def app_factory(api_name: str = "api", vocab_route: str = "vocab") -> Flask:
     CORS(app)
     app.url_map.strict_slashes = False
     api = Api(app)
+
+    @app.after_request
+    def logging_request_response(response):
+        logger = HydrusLogger().get_logger()
+        logger.info("{} {}/<iri>".format(uuid.uuid4(), request.method))
+        return response
 
     # Redirecting root_path to root_path/api_name
     if api_name:
--- a/hydrus/app_factory.py
+++ b/hydrus/app_factory.py
@@ -5,7 +5,7 @@ from flask_restful import Api
 from hydrus_logger import HydrusLogger
 from hydrus.resources import (Index, Vocab, Contexts, Entrypoint,
                               ItemCollection, Item, Items, ItemMember)
-
+logger = HydrusLogger().get_logger()
 
 def app_factory(api_name: str = "api", vocab_route: str = "vocab") -> Flask:
     """
@@ -21,9 +21,9 @@ def app_factory(api_name: str = "api", vocab_route: str = "vocab") -> Flask:
     app.url_map.strict_slashes = False
     api = Api(app)
 
+
     @app.after_request
     def logging_request_response(response):
-        logger = HydrusLogger().get_logger()
         logger.info(" {} {}:{} Object_created_at : {}, status_code : {} "
                     .format(uuid.uuid4(), request.method, request.path, response.location, response.status))
         return response
--- a/hydrus/app_factory.py
+++ b/hydrus/app_factory.py
@@ -5,7 +5,9 @@ from flask_restful import Api
 from hydrus_logger import HydrusLogger
 from hydrus.resources import (Index, Vocab, Contexts, Entrypoint,
                               ItemCollection, Item, Items, ItemMember)
+
 logger = HydrusLogger().get_logger()
+
 
 def app_factory(api_name: str = "api", vocab_route: str = "vocab") -> Flask:
     """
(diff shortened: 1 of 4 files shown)


farazkhanfk7:
Works fine. Except for one issue that this was working fine when `fileConfig` was being used instead of `dictConfig`. Now it's not creating `hydrus.log` file after the latest commit. Please cross-check and let me know if I made any mistake here. I think we have to use relative path of LogConfig in HydrusLogger class. 

Also I can see, that `resource_url` is not being fetched yet and is hardcoded to `<iri>`. 
Suggestion : 
- You can use `request.path` to get the complete URL of path.
- Also to get the IRI of newly created/updated object you can add conditional checks for `response.status_code` and return IRI of object using `response.location`.

I trying adding these on my local. This was the format after creation of a new object:
```
2021-05-15 11:52.10 [info     ] 327bbe6e-0280-4eaf-99a3-0222e52fc170 GET:/serverapi/MovieCollection, Status Code : [200 OK]
2021-05-15 11:53.23 [info     ] 29dcabbe-1c46-48b1-bedc-b9913c24afdb PUT:/serverapi/Movie/,  Object created with ID:http://localhost:8080/serverapi/Movie/0134ee40-a090-484f-97ee-df41d993e775
```

hydrus/LogConfig.py
```diff
@@ -0,0 +1,27 @@
+from os.path import abspath, dirname, join
```
Mec-iS: name of modules is lowercase and use underscores. please be attentive to previous decisions and maintain coherence in the codebase

hydrus/app_factory.py
```diff
@@ -20,6 +21,13 @@ def app_factory(api_name: str = "api", vocab_route: str = "vocab") -> Flask:
     app.url_map.strict_slashes = False
     api = Api(app)
 
+    @app.after_request
+    def logging_request_response(response):
+        logger = HydrusLogger().get_logger()
```
Mec-iS: definition of logger should be done at module-level, not inside a function.

This line should be at the top of the module, after imports.

Mec-iS (changes requested):
as mentioned in a comment, the log level cannot be hardcoded but needs to be set according to an environment variable, please be more attentive when reading suggestions.

`level` should get the value from Python `environ` or default to `DEBUG` if the value is null.

hydrus/LogConfig.py
Purvanshsingh: This file is removed. we are not using `filedictConfig()` instead we are using `fileConfig('LogConfig')`

hydrus/app_factory.py
Purvanshsingh: > as mentioned in a comment, the log level cannot be hardcoded but needs to be set according to an environment variable, please be more attentive when reading suggestions.
> 
> `level` should get the value from Python `environ` or default to `DEBUG` if the value is null.

I just made it hardcoded for testing and verification purpose.

Purvanshsingh:
> as mentioned in a comment, the log level cannot be hardcoded but needs to be set according to an environment variable, please be more attentive when reading suggestions.
> 
> `level` should get the value from Python `environ` or default to `DEBUG` if the value is null.

I just made it hardcoded for testing & Verification purpose.

hydrus/LogConfig.py
Purvanshsingh: We have `LogConfig` file for configuring the logging object.

Purvanshsingh:
@Mec-iS  Can we set the default logging level to `INFO` instead of `DEBUG` as we are using environment variable all the logging messages are being logged to the file.

Especially` webscoket.server` logging messages.
Screenshot from 2021-05-15 15-45-57

hydrus/app_factory.py
```diff
@@ -20,6 +21,13 @@ def app_factory(api_name: str = "api", vocab_route: str = "vocab") -> Flask:
     app.url_map.strict_slashes = False
     api = Api(app)
 
+
+    @app.after_request
+    def logging_request_response(response):
+        logger.info(" {} {}:{} Object_created_at : {}, status_code : {} "
```
farazkhanfk7: I think this line should be added inside an if-else block ( `if response.location : `) to check if we're getting a `response.location` or not. Otherwise `Object_created_at :` will keep returning `None` even in case of GET requests to other endpoints. In that case, we should only log the method and path of request.

Mec-iS:
> @Mec-iS Can we set the default logging level to `INFO` instead of `DEBUG` as we are using environment variable all the logging messages are being logged to the file.
> 
> Especially` webscoket.server` logging messages.

those are relevant logs, there should be a way to wrap those logs in the structlog context as well.

In general, the project is still in debugging phase so the right level is `DEBUG`. Once we move closer to a production-ready system it will be more correct to set it to `INFO`.

hydrus/app_factory.py
Purvanshsingh: @farazkhanfk7 I have done that, just waiting for the level `DEBUG` approval to push all the changes.
Thanks.

Mec-iS: please see my comment below

Purvanshsingh: @Mec-iS yes, I am figuring out the way to deal with those logs with structlog.

Purvanshsingh:
> > @Mec-iS Can we set the default logging level to `INFO` instead of `DEBUG` as we are using environment variable all the logging messages are being logged to the file.
> > Especially` webscoket.server` logging messages.
> 
> those are relevant logs, there should be a way to wrap those logs in the structlog context as well.
> 
> In general, the project is still in debugging phase so the right level is `DEBUG`. Once we move closer to a production-ready system it will be more correct to set it to `INFO`.

@Mec-iS I went through the structlog context, we can only structure those messages to any format that we have logged.
I tried doing that as well but these messages are from ` webscoket.server` they can't be structured. they are being logged because we are using `DEBUG` level they won't be logged as we change to `INFO`.

Mec-iS:
leave them there for now

sameshl (approved):
Looks good. Great work! :+1:

Mec-iS (approved):

PR #593, opened 2021-06-25
Added foreign key relationship between a Collection and managed Class

### LINKED ISSUE #592: Foreign key relationship between Collection and managed Class
### I'm submitting a 
- [ ] bug report.
- [x] feature request.

### Current Behaviour:

Presently, if an object is deleted from a Class it would still be there in the collection table in database( if it was previously inserted in a Collection that directly manages that Class). In such case, a GET response from Collection endpoint will show the member available in the database but will throw an InstanceNotFound error when a GET request will be made to view details of the object because it has already been deleted from the Class.

### Expected Behaviour:

hydrus should check if an object is also present in the database table of a Collection that directly manages the class and then delete the object from both concurrently. There should be foreign-key relationship between a Class and a Collection that manages that Class. We can use `ondelete=CASCASE` and `onupdate` while creating a foreign key column. Using this, a member could be deleted automatically if it’s available in it’s Class table.

### Environment:
 
* python version
* pip version
* OS details

farazkhanfk7:
Fixes #592

### Checklist

### Current behaviour

Presently, if an object is deleted from a Class it would still be there in the collection table in database( if it was previously inserted in a Collection that directly manages that Class). This was because there was no foreign key relationship between a Class and a Collection. 

### New expected behaviour

image
_img : schema -> ID of a class object will be a foreign key to `member` in a Collection_ 
### Change logs
-  Change `member` in Collection table to a foreign key column
-  Added test to check members from a Collection are deleted after the class objects are deleted from the database.

### DIFF
--- a/hydrus/data/db_models.py
+++ b/hydrus/data/db_models.py
@@ -111,6 +111,9 @@ class Resource:
             )
             # if the class is a collection, add an extra column for
             # member @type
+            manages = self.resource["manages"]
+            managed_class = manages["object"].split(expanded_base_url)[1]
+            attr_dict[title] = Resource.foreign_key_column(managed_class, title)
             attr_dict["member_type"] = Column(
                 String,
             )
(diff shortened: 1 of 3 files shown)


Mec-iS (approved):

Mec-iS:
we have to remove the Travis branch check as it is now useless and never starts

PR #594, opened 2021-07-03
Added support for datatype for columns in database tables.

### LINKED ISSUE #581: Define and implement a way to set right column types when auto-generating the database
### I'm submitting a 
- [x ] feature request.

### Current Behaviour:
All database columns are set to `db.String` as Hydra do not require specifying the type of variable for a filed

### Expected Behaviour:
Database columns should be created with `dbString`, `db.Integer` or `db.Float` according to types defined in the ApiDoc or Ontology via properties like `"@type": " "xsd:decimal"` as discussed in https://github.com/HTTP-APIs/creditrisk-poc/issues/6

I can work on this but I need support from @sameshl

farazkhanfk7:
Fixes #581 

### Checklist

### Current behaviour

Database columns should be created with String, Integer or Float according to types defined in the ApiDoc.
### New expected behaviour

hydrus will check for `range` of supported_properties. If the supported_property has a `range` ( specified datatype ) then column in database table will be created accordingly. For ex: 
```
"range": "https://www.w3.org/TR/xmlschema-2/#integer"
```
Currently, I have added support for `Integer` and `Float` only. We can also add `DateTime` column after this. If a user won't pass `range` as an argument while creating `HydraClassProp` then the column will be treated as String in hydrus database.
 Example (Passed "xsd:integer" as range for two random HydraClassProp in apidoc):
img

Also, I've made changes in `gen_dummy_object` function, so that it'll also check the type of column before creating it's dummy object. If the column type is integer or float, then a random integer is returned ( instead of a string).
### Change logs

- Made Changes in db_models.py
- Made changes in gen_dummy_object function.
- Updated the existing docs in hydrus

### DIFF
--- a/examples/drones/doc.py
+++ b/examples/drones/doc.py
@@ -58,7 +58,8 @@ doc = {
         "supportedOperation": "hydra:supportedOperation",
         "supportedProperty": "hydra:supportedProperty",
         "title": "hydra:title",
-        "writeable": "hydra:writeable"
+        "writeable": "hydra:writeable",
+        "xsd": "https://www.w3.org/TR/xmlschema-2/#"
     },
     "@id": "http://localhost:8080/api/vocab",
     "@type": "ApiDocumentation",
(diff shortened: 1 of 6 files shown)


farazkhanfk7:
@Mec-iS  I think checks are failing because Travis is using a different version of hydra-python-core which hasn't been released yet. The job throws error that says that HydraClassProp object doesn't have a kwargs attribute. The tests are passing on my local. Can u please confirm.

Mec-iS:
ok. let's release the new hydra-python-core and update the requirements.
please @farazkhanfk7 create an account at https://pypi.org/ and send me your username on Slack

sameshl:
> If the column type is integer or float, then a random integer is returned

@farazkhanfk7 If the column type is a float, shouldn't a float value be returned?

farazkhanfk7:
@sameshl  Yes, a float should be returned. But if the column type is Float, an integer 20 will be treated as 20.0 in database. This is why I created a random integer for both types.

Mec-iS (approved):

PR #599, opened 2021-07-26
Added support for datetime column in datebase created by hydrus

### LINKED ISSUE #598: Add support for datetime column in hydrus
### I'm submitting a 
- [ ] bug report.
- [x] feature request.

### Current Behaviour:

Presently, we can create columns in database for only integer, float, string datatypes in database. 
### Expected Behaviour:

A user should be able to specify datatype like `"xsd:dateTime"` while creating API doc. Database columns should be created accordingly. Ref  : https://www.w3.org/TR/xmlschema-2/#dateTime

### Environment:
 
* python version
* pip version
* OS details

Yes

farazkhanfk7:
Fixes #598 

### Checklist

### Current behaviour

Previously, we can had support for only integer, float, string datatypes in database. I've added support for a Datetime column.

### New expected behaviour

- SQLAlchemy Datetime Column only takes a datetime object. But a user cannot send a datetime object in request body ( not json serializable ). This is why I created helper functions to convert datetime string to a python datetime object and then it'll be inserted to database.
- The helper function (`get_modified_object`)  checks if there is any supported property in object (request body) which is actually a datetime field. And then returns a modified object accordingly.
- The helper function needs a doc to check which supported_properties are actually datetime fields in an object. The only problem was crud tests were failing because crud.insert is called directly in these tests. Couldn't call `get_doc()` directly in `crud.insert_object` So I had to pass the doc as an argument in `crud.insert` and `crud.insert_object` functions as well. 
- Everything is tested, tests are passing. Also to test that columns are created with correct datatype, I've made required changes in _conftest.py_'s `gen_dummy_object` function.
-  Added datatype for Prop1 and Prop2 supported properties in doc_writer_sample in hydrus. ( It is used for testing )
### Change logs 
- db_models.py
- resource_based_classes.py
- item_collection_helpers.py
- itemhelpers.py
- items_helpers.py
- crud.py
- conftest.py
- test_crud.py
- doc_writer_sample.py

### DIFF
--- a/hydrus/data/crud.py
+++ b/hydrus/data/crud.py
@@ -97,6 +97,7 @@ def get(
 
 
 def insert(
+    doc_,
     object_: Dict[str, Any],
     session: scoped_session,
     id_: Optional[str] = None,
@@ -124,12 +125,12 @@ def insert(
     object_template = copy.deepcopy(object_)
     if id_ is not None:
         object_template["id"] = id_
-    inserted_object_id = insert_object(object_template, session, collection)
+    inserted_object_id = insert_object(doc_, object_template, session, collection)
     return inserted_object_id
 
 
 def insert_multiple(
-    objects_: List[Dict[str, Any]], session: scoped_session, id_: Optional[str] = ""
+    doc, objects_: List[Dict[str, Any]], session: scoped_session, id_: Optional[str] = ""
 ) -> List[str]:
     """
     Adds a list of object with given ids to the database
@@ -167,7 +168,7 @@ def insert_multiple(
             pass
         except TypeError:
             pass
-        inserted_object_id = insert(object_, session, id_of_object_)
+        inserted_object_id = insert(doc, object_, session, id_of_object_)
         instance_id_list.append(inserted_object_id)
 
     return instance_id_list
@@ -209,6 +210,7 @@ def delete_multiple(id_: List[int], type_: str, session: scoped_session) -> None
 
 
 def update(
+    doc_,
     id_: str,
     type_: str,
     object_: Dict[str, str],
@@ -228,7 +230,7 @@ def update(
     :return: id of updated object
     """
     query_info = {"@type": type_, "id_": id_}
-    updated_object_id = update_object(object_, query_info, session, collection)
+    updated_object_id = update_object(doc_, object_, query_info, session, collection)
     return updated_object_id
 
 
--- a/hydrus/data/crud.py
+++ b/hydrus/data/crud.py
@@ -49,6 +49,7 @@ from hydrus.data.crud_helpers import (
 
 # from sqlalchemy.orm.session import Session
 from sqlalchemy.orm.scoping import scoped_session
+from hydra_python_core.doc_writer import HydraDoc
 from typing import Dict, Optional, Any, List
 
 from hydrus.data.resource_based_classes import (
@@ -97,13 +98,14 @@ def get(
 
 
 def insert(
-    doc_,
+    doc_: HydraDoc,
     object_: Dict[str, Any],
     session: scoped_session,
     id_: Optional[str] = None,
     collection: bool = False,
 ) -> str:
     """Insert an object to database [POST] and returns the inserted object.
+    :param doc_ : Hydra Doc object
     :param object_: object to be inserted
     :param session: sqlalchemy scoped session
     :param id_: id of the object to be inserted (optional param)
@@ -130,10 +132,14 @@ def insert(
 
 
 def insert_multiple(
-    doc, objects_: List[Dict[str, Any]], session: scoped_session, id_: Optional[str] = ""
+    doc: HydraDoc,
+    objects_: List[Dict[str, Any]],
+    session: scoped_session,
+    id_: Optional[str] = ""
 ) -> List[str]:
     """
     Adds a list of object with given ids to the database
+    :param doc : Hydra Doc object
     :param objects_: List of dict's to be added to the database
     :param session: scoped session from getSession in utils
     :param id_: optional parameter containing the ids of objects that have to be inserted
@@ -210,7 +216,7 @@ def delete_multiple(id_: List[int], type_: str, session: scoped_session) -> None
 
 
 def update(
-    doc_,
+    doc_: HydraDoc,
     id_: str,
     type_: str,
     object_: Dict[str, str],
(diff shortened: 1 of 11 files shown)


hydrus/data/crud.py
```diff
@@ -97,6 +97,7 @@ def get(
 
 
 def insert(
+    doc_,
```
Mec-iS: new `doc_` element not in docstring. Please update the docstring

hydrus/data/resource_based_classes.py
```diff
@@ -47,7 +74,7 @@ def get_database_class(type_: str):
 
 
 def insert_object(
-    object_: Dict[str, Any], session: scoped_session, collection: bool = False
+    doc_, object_: Dict[str, Any], session: scoped_session, collection: bool = False
```
Mec-iS: new `doc_` element not in docstring

hydrus/data/resource_based_classes.py
```diff
@@ -34,6 +36,31 @@ def get_type(object_: Dict[str, Any]) -> str:
...
+        title = getattr(prop, 'title')
+        if range is not None:
+            if "dateTime" in range:
+                datetimefield = title
+                datetimefields.append(datetimefield)
+    if len(datetimefields) != 0:
+        for field in datetimefields:
+            try:
+                datetime_value = temp_object.get(field)
+                dt_object = datetime.strptime(datetime_value, "%d/%m/%Y %H:%M:%S")
```
Mec-iS: Please use standard tools to check for string compliance. See [this answer](https://stackoverflow.com/a/61569783/2536357)

hydrus/data/resource_based_classes.py
```diff
@@ -198,6 +229,7 @@ def delete_object(
 
 
 def update_object(
+    doc_,
```
Mec-iS: `doc_` missing in docstring

tests/conftest.py
```diff
@@ -89,7 +91,15 @@ def gen_dummy_object(class_title, doc):
                 else:
                     type_ = prop.kwargs.get('range')
                     if type_ is not None:
-                        object_[prop.title] = random.randint(50,100)
+                        if "dateTime" in type_:
+                            value = datetime.utcnow()
+                            obj = datetime.strftime(value, "%d/%m/%Y %H:%M:%S")
```
Mec-iS: same as above. you should avoid using hard-coded arbitrary formats for date-times, we need to stick to conventiion that in the datetime case is the [ISO 8601 format](https://www.cl.cam.ac.uk/~mgk25/iso-time.html).
check [this post](https://stackoverflow.com/questions/127803/how-do-i-parse-an-iso-8601-formatted-date) about the timezone information: timezone in the database has to be always UTC, the client should pass the datetime always with the timezone specification (so it has to be checked by the server). The ideal ISO 8601 format in my experience is `2008-09-03T20:56:35.450686Z`, see the `T` separator between date and time and the indication of the timezone `Z`

Mec-iS (changes requested):

Mec-iS:
> The helper function needs a doc to check which supported_properties are actually datetime fields in an object. The only problem was crud tests were failing because crud.insert is called directly in these tests. Couldn't call get_doc() directly in crud.insert_object So I had to pass the doc as an argument in crud.insert and crud.insert_object functions as well.

@chrizandr @sameshl @priyanshunayan please check that this the right imlementation

hydrus/data/resource_based_classes.py
Mec-iS: also `doc_` is missing the type hint

farazkhanfk7:
Thanks for pointing that out. I missed docstring and type annotations. 
I'll add that.

hydrus/data/resource_based_classes.py
```diff
@@ -60,7 +60,7 @@ def get_modified_object(object_: Dict[str, Any], doc: HydraDoc, path: str) -> Di
         for field in datetimefields:
             try:
                 datetime_value = temp_object.get(field)
-                dt_object = datetime.strptime(datetime_value, "%d/%m/%Y %H:%M:%S")
+                dt_object = parser.isoparse(datetime_value)
```
Mec-iS, line 63: what happens if the value passed by the client is not ISO and where we return a 400 error if that happens?

farazkhanfk7: Sure. I'll raise an exception here. I think in this case it'll be better to create a new custom exception to raise HydraError.

farazkhanfk7: Added a custom exception which can be raised if the input value for datetime property is not in ISO format. A 400 Response is returned:
```
{
    "@context": "https://www.w3.org/ns/hydra/core",
    "@type": "Error",
    "description": "The format of Prop1 is invalid. Datetime input should be in ISO format.",
    "statusCode": 400,
    "title": "Invalid Datetime format"
}
```

hydrus/data/resource_based_classes.py
```diff
@@ -34,6 +37,40 @@ def get_type(object_: Dict[str, Any]) -> str:
...
 
+def get_modified_object(object_: Dict[str, Any], doc: HydraDoc, path: str) -> Dict[str, Any]:
+    """
+    Return the object after modifying properties.
+    :param object_: Dict containing object properties
+    :param doc: HydraDoc object
+    :param path: Path of the Class or Collection
+    :return: object_ containing properties
+    """
+    temp_object = copy.deepcopy(object_)
```
Mec-iS: why you need to copy this object? this is a very expensive operation

farazkhanfk7: oops ! I was using this in a previous approach. Thanks for catching that.

hydrus/data/resource_based_classes.py
```diff
@@ -34,6 +37,39 @@ def get_type(object_: Dict[str, Any]) -> str:
...
+    if len(datetimefields) != 0:
+        for field in datetimefields:
+            try:
+                datetime_value = object_.get(field)
+                dt_object = parser.isoparse(datetime_value)
+                object_[field] = dt_object
+            except ValueError:
+                raise InvalidDateTimeFormat(field)
+            except TypeError:
+                datetime_value = object_.get(field)
```
Mec-iS, line 68: please add a comment here to specify which error case this exception should handle as it is not clear

Mec-iS (approved):

PR #611, opened 2022-02-23
Fixing typo in `demo.py` file

### LINKED ISSUE #610: Typo in `demo.py` file
### I'm submitting a 
- [x] bug report.

### Current Behaviour:
While executing the [demo.py](https://github.com/HTTP-APIs/hydrus/blob/develop/examples/hydrus-demo-server/demo.py) file, it throws the following error: `from gevent.wsgi import WSGIServer
ModuleNotFoundError: No module named 'gevent.wsgi'`

### Expected Behaviour:
The server should start running

### Steps to reproduce:
Run the [demo.py](https://github.com/HTTP-APIs/hydrus/blob/develop/examples/hydrus-demo-server/demo.py) file.

### Snapshot:
image

### Environment:
 
* python version - 3.10.2
* pip version - 22.0.3
* OS details
    - Name - Microsoft Windows 11 Home Single Language
    - Version - 10.0.22000 N/A Build 22000

Yes

vrutik2809:
Fixes #610 

### Checklist

### Current behaviour
There is a typo in the `demo.py` file. Due to that the following error was occurring:
image

### New expected behaviour
The typo has been removed. Now it will work completely fine.

#### Fixed 

- Typo removed

### DIFF
--- a/examples/hydrus-demo-server/demo.py
+++ b/examples/hydrus-demo-server/demo.py
@@ -10,7 +10,7 @@ from hydra_python_core import doc_maker
 from hydrus.data.db_models import Base
 from hydrus.data.user import add_user
 from doc import doc
-from gevent.wsgi import WSGIServer
+from gevent.pywsgi import WSGIServer
 
 
 if __name__ == "__main__":
--- a/examples/hydrus-demo-server/demo.py
+++ b/examples/hydrus-demo-server/demo.py
@@ -64,7 +64,7 @@ if __name__ == "__main__":
     with set_authentication(app, True):
         # Use authentication for all requests
         with set_token(app, True):
-            with set_api_name(app, "serverapi"):
+            with set_api_name(app, API_NAME):
                 # Set the API Documentation
                 with set_doc(app, apidoc):
                     # Set HYDRUS_SERVER_URL
--- a/examples/hydrus-demo-server/demo.py
+++ b/examples/hydrus-demo-server/demo.py
@@ -56,10 +56,6 @@ if __name__ == "__main__":
     print("Adding authorized users")
     add_user(id_=1, paraphrase="test", session=session)
 
-    # Insert them into the database
-    doc_parse.insert_classes(classes, session)
-    doc_parse.insert_properties(properties, session)
-
     print("Creating the application")
     # Create a Hydrus app with the API name you want, default will be "api"
     app = app_factory(API_NAME)


Mec-iS:
please check if the same declaration is present in other parts of the codebase

Mec-iS (approved):

vrutik2809:
### Another two issues found in the same file:
- No method named `insert_classes` in `doc_parse.py` module (line: 60)
- No method named `insert_properties` in `doc_parse.py` module (line: 61)

Need to remove those lines

Mec-iS:
you should provide the changes to fix those then. Are those methods unused or they changed names? Check `hydra-python-core` repository also.

Mec-iS:
Those methods have been removed from `hydrus/data/doc_parse.py` so `demo.py` should be updated.

Mec-iS:
@farazkhanfk7 do you know what we use now in place of `insert_classes` to generate the class objects?

farazkhanfk7:
In `hydrus/data/db_models.py`, there is a Resource class which is used to create a Class or Collection. It has a function `make_db_table`, which is called by another function `create_database_tables` to insert classes in the database.
https://github.com/HTTP-APIs/hydrus/blob/c6a8587c7904afe64c31f74da1f9d4edc7139ed2/hydrus/data/db_models.py#L217-L220

vrutik2809:
Unavailable method removed from `demo.py` file

Mec-iS:
ok. but now the demo should work by using the new methods.

farazkhanfk7:
@vrutik2809 you can take a look at `app.py` for reference and also make sure that hydrus server is working fine after running `demo.py`.

vrutik2809:
While running `demo.py`, it is running two times:
- First, when I run `demo.py` it's running on `http://localhost:8080/api` without going in the `if` block
image

- Second when I hit `Ctrl + c` (in order to stop the server) then it actually goes into the `if` block and starts running on `http://localhost:8080/serverapi` but it shows `404 not found`
image

I'm not able to figure out why it is happening.

While debugging it, I also observed that even if pass a different `port` or `api-name` it still listen to  `base-url & api-name` provided while making documentation using `doc.writer.HydraDoc()`. It can be verified in the above image also. Api-name in the `doc.py` (reference image attached below) is `api` but in `demo.py` it is passed as `serverapi`  then it is throwing `404` but when I fire a request on `/api` then it gives me `200` 

snap of [doc.py](https://github.com/HTTP-APIs/hydrus/blob/develop/examples/hydrus-demo-server/doc.py) :
image

vrutik2809:
I have opened new issue ([#613](https://github.com/HTTP-APIs/hydrus/issues/613)) regarding `PORT` and `API_NAME` problem.

Mec-iS (approved):

Mec-iS:
Thanks!

